PR #13, opened 2021-04-12
Add slurm prolog epilog

jamesbeedy:
These changes add a prolog and epilog for slurmctld along with associated files.


jamesbeedy:
needs tests

PR #20, opened 2021-05-18
V1 additional dev david

djkresta:
https://omnivector-solutions.monday.com/boards/1204470680/pulses/1299676711

These changes introduce enhanced license tracking in the slurmctld
prolog that enforce only tracking licenses that meet the protocol:
`<product>.<feature>@<license_server>:<num_tokens>`

* test that no licenses returned when correct protocol not used
* test that correct license count is returned when correct
protocol is used
* move call to scontrol and output parsing to cmd_utils.py
* enhance slurmctldProlog to only track licenses that match the
protocol

### DIFF
--- a/src/licensemanager2/test/agent/test_slurmctld_prolog.py
+++ b/src/licensemanager2/test/agent/test_slurmctld_prolog.py
@@ -1,0 +1,48 @@
+"""
+Tests of Prolog
+"""
+
+from typing import List
+# from unittest.mock import patch
+from pytest import fixture, mark
+
+
+@fixture
+def scontrol_parsed_output_good() -> List[str]:
+    return [
+        "TESTPRODUCT.TESTFEATURE@flexlm:11",
+        "TESTPRODUCT2.TESTFEATURE2@flexlm:22",
+        "TESTPRODUCT3.TESTFEATURE3@flexlm:33",
+    ]
+
+
+@fixture
+def scontrol_parsed_output_bad() -> List[str]:
+    return [
+        "TESTFEATURE@flexlm:11",
+        "TESTPRODUCT2@22",
+        "TESTFEATURE3@flexlm",
+    ]
+
+
+@mark.asyncio
+async def test_get_required_licenses_for_job():
+    """
+    Do I collect the requested structured data from running all these dang tools?
+    """
+    p0 = patch.object(
+        cmd_utils,
+        "scontrol_show_lic",
+        scontrol_output_good_format
+    )
+    with p0:
+        licenses = await slurmctld_prolog._get_required_licenses_for_job()
+        assert len(licenses) == 3
+    
+
+@mark.asyncio
+async def test_get_required_licenses_for_job():
+    """
+    Do I collect the requested structured data from running all these dang tools?
+    """
+    assert (0 == 1)
--- a/src/licensemanager2/test/agent/test_slurmctld_prolog.py
+++ b/src/licensemanager2/test/agent/test_slurmctld_prolog.py
@@ -1,10 +1,13 @@
 """
 Tests of Prolog
 """
-
 from typing import List
 # from unittest.mock import patch
 from pytest import fixture, mark
+from unittest.mock import patch
+from licensemanager2.workload_managers.slurm.slurmctld_prolog import (
+    get_required_licenses_for_job,
+)
 
 
 @fixture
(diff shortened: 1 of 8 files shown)


src/licensemanager2/test/agent/test_slurmctld_prolog.py
```diff
@@ -0,0 +1,65 @@
+"""
+Tests of Prolog
+"""
+from typing import List
+# from unittest.mock import patch
```
zsinx6: This line can be removed

src/licensemanager2/test/agent/test_slurmctld_prolog.py
```diff
@@ -0,0 +1,65 @@
...
+    Do I return the correct licenses when the license format matches?
+    """
+    p1 = patch(
+        'licensemanager2.workload_managers.slurm.slurmctld_prolog.get_licenses_for_job',
+        return_value=scontrol_parsed_output_good
+    )
+    with p1:
+        license_booking_request = await get_required_licenses_for_job(
+            slurm_job_id
+        )
```
zsinx6: What you think about changing this to use the patch as a test function decorator? In that way you can avoid indentation and will be easier to read.
We are already using this here, for example: https://github.com/omnivector-solutions/jobbergate-api-fastapi/blob/dac55d996930e0dcc7c584842c5e9e857438e562/jobbergateapi2/tests/apps/job_scripts/test_routers.py#L104

src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -14,6 +14,42 @@
...
+        CMD_TIMEOUT
+    )
+    scontrol_out = str(scontrol_out, ENCODING)  # type: ignore
+    logger.info(scontrol_out)
+
+    # Check that the command completed successfully
+    if not scontrol_show_lic.returncode == 0:
+        msg = f"Could not get SLURM data for job id: {slurm_job_id}"
+        logger.error(msg)
+        raise Exception(msg)
```
zsinx6: It is a good practice to create our own class as subclass of Exception with a good name and docstring, just raising Exception it bad for readability.

```python
class MyException(Exception):
"""This exception is used when..."""
```

tox.ini
```diff
@@ -15,11 +15,11 @@ whitelist_externals = poetry
 
 [testenv:mypy]
 commands =
-    poetry run mypy {posargs} src/licensemanager2/ jawthorizer/ --exclude src/licensemanager2/workload_managers
+    poetry run mypy {posargs} src/licensemanager2/ jawthorizer/
 
 [testenv:lint]
 commands =
-    poetry run flake8 {posargs} --ignore=D400,W605,W503 src/licensemanager2/ jawthorizer/
+    poetry run flake8 {posargs} --ignore=W292,D400,W605,W503 src/licensemanager2/ jawthorizer/
```
zsinx6, line 22: Why ignore the W292?

zsinx6 (changes requested):

zsinx6:
And about the PR title, it is good to be a brief summary about what have been changed, this one name says nothing about the PR for someone outside of the project.

src/licensemanager2/workload_managers/slurm/slurmctld_prolog.py
```diff
@@ -223,7 +185,8 @@ async def main():
     sys.exit(0)
 
 
-# Initialize the logger
-init_logging("slurmctld-prolog")
-# Run main()
-asyncio.run(main())
+if __name__ == "__main__":
+    # Initialize the logger
+    init_logging("slurmctld-prolog")
```
heitorPB: I'm not sure `uvicorn` will call the `init_logging` function this way. Normally we initialize the logging system in the beginning of the file.

jamesbeedy:
> And about the PR title, it is good to be a brief summary about what have been changed, this one name says nothing about the PR for someone outside of the project.

I agree with this, and would also like to see a link to the monday item that drives this work in the description when possible.

```
Fixes: https://<url-to-monday-item>
```

tox.ini
djkresta: Ignored because per James we don't want a newline at the end of our file.

src/licensemanager2/workload_managers/slurm/slurmctld_prolog.py
djkresta: We had to put them inside main so they wouldn't be run at import time, but since figured out a better solution: move the functions out of the prolog into cmd_utils.py

src/licensemanager2/test/agent/test_slurmctld_prolog.py
djkresta: Sorry, I see you are talking about the commented out line.. removed.

src/licensemanager2/test/agent/test_slurmctld_prolog.py
djkresta: I am not quite clear on how you are setting up the return value in the example you linked. From examples I have constructed this:
```
@mock.patch("licensemanager2.workload_managers.slurm.slurmctld_prolog.get_licenses_for_job", return_value=scontrol_parsed_output_good)
async def test_get_required_licenses_for_job_good(slurm_job_id: str, scontrol_parsed_output_good: List[str]):
    """
    Do I return the correct licenses when the license format matches?
    """
    license_booking_request = await get_required_licenses_for_job(
        slurm_job_id
    )
    assert len(license_booking_request.bookings) == 3
```

but I get:
```
>       if license_array[0] != "(null)":
E       TypeError: 'function' object is not subscriptable

src/licensemanager2/workload_managers/slurm/slurmctld_prolog.py:62: TypeError
```

I see in your example you set the return value in the function using 
`boto3_client_mock.client.return_value = s3_client_mock` 
but I don't see how that connects with the decorator
`@mock.patch("jobbergateapi2.apps.job_scripts.routers.boto3")`

Because boto3_client_mock is passed into the function as an argument.

src/licensemanager2/test/agent/test_slurmctld_prolog.py
zsinx6:
```suggestion
```

zsinx6 (approved):

src/licensemanager2/workload_managers/slurm/slurmctld_prolog.py
```diff
@@ -11,175 +11,25 @@
...
 from licensemanager2.agent import log, init_logging
-from licensemanager2.workload_managers.slurm.common import (
-    LM2_AGENT_HEADERS,
-    SCONTROL_PATH,
-    CMD_TIMEOUT,
-    get_job_context
+from licensemanager2.workload_managers.slurm.common import get_job_context
+from licensemanager2.workload_managers.slurm.cmd_utils import (
+    LicenseBookingRequest,
+    _force_reconciliation,
```
jamesbeedy: @djkresta make these functions public

jamesbeedy:
👍

PR #24, opened 2021-05-25
Fix epilog script

djkresta:
These changes fix a problem where we were iterating over license_booking_request instead of license_booking_request.bookings.
Also await for get_tokens_for_license.


jamesbeedy:
👍

PR #26, opened 2021-06-03
info -> debug and log moar

jamesbeedy:
These changes introduce enhanced logging in the prolog and epilog
along with log level modifications that ensure we arent logging
so many things to the info log. Additionally, increase the size of
the running log to 5MB before rolling it.
[monday story](https://omnivector-solutions.monday.com/boards/1204470680/pulses/1299704236)

### DIFF
--- a/src/licensemanager2/agent/__init__.py
+++ b/src/licensemanager2/agent/__init__.py
@@ -55,8 +55,8 @@ def _rotating_file_handler(log_file: str):
 
     handler = RotatingFileHandler(
         log_file,
-        backupCount=5,
-        maxBytes=100000,
+        backupCount=10,
+        maxBytes=5242880,
     )
     handler.rotator = _rotator
     handler.namer = _namer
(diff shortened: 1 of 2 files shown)


djkresta:
The Monday item says something about logging the job_id, and I don't see that added. Do we need to add that info to the logs? Also the tests are not passing right now.

src/licensemanager2/agent/__init__.py
```diff
@@ -55,8 +55,8 @@ def _rotating_file_handler(log_file: str):
 
     handler = RotatingFileHandler(
         log_file,
-        backupCount=5,
-        maxBytes=100000,
+        backupCount=10,
+        maxBytes=5242880,
```
djkresta, line 60: could probably use a comment, and also this is not captured in the monday story. (mention that we are also increasing backup size)

jamesbeedy: ahh good catch - I've updated the monday story and added a comment around the log size.

jamesbeedy:
> The Monday item says something about logging the job_id, and I don't see that added. Do we need to add that info to the logs? Also the tests are not passing right now.

Good catch here. I've updated the logging to use the job_id where appropriate.

djkresta (approved):
Those changes look good!

PR #35, opened 2021-07-19
Fixed mypy issues with cmd_utils

### LINKED ISSUE #19: mypy errors with subprocess output types
When running mypy these type errors were shown around subprocess output and regex types. TODO: fixme (we silenced the errors for now)
```
src/licensemanager2/workload_managers/slurm/cmd_utils.py:32: error: Incompatible types in assignment (expression has type "str", variable has type "bytes")
src/licensemanager2/workload_managers/slurm/cmd_utils.py:42: error: Value of type variable "AnyStr" of "search" cannot be "object"
src/licensemanager2/workload_managers/slurm/cmd_utils.py:43: error: Item "None" of "Optional[Match[object]]" has no attribute "group"
src/licensemanager2/workload_managers/slurm/cmd_utils.py:43: error: Item "object" of "Union[object, Any]" has no attribute "split"
```

dusktreader:
Fixed the mypy issues with typing in the get_licenses_for_job method.
Also added 3 unit tests for the method.

Fixes #19

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,7 +79,7 @@ dev = [
     "uvicorn",
     "pydantic",
     "alembic",
-    ]
+]
 
 unit = [
     "aiosqlite",
@@ -96,7 +96,7 @@ unit = [
     "sqlalchemy-stubs",
     "pydantic",
     "virtualenv",
-    ]
+]
 
 mypy = [
     "mypy",
@@ -106,7 +106,12 @@ mypy = [
     "types-Jinja2",
     "types-jwt",
     "types-click",
-    ]
+    "sqlalchemy-stubs",
+    "httpx",
+    "pytest",
+    "respx",
+    "aiosqlite",
+]
 
 
 
--- a/src/licensemanager2/test/agent/test_slurmctld_prolog.py
+++ b/src/licensemanager2/test/agent/test_slurmctld_prolog.py
@@ -58,7 +58,7 @@ async def test_get_licenses_for_job_good(
         JobId=2 JobName=sleep
            UserId=root(0) GroupId=root(0) MCS_label=N/A
            Priority=4294901759 Nice=0 Account=root QOS=normal
-           JobState=CANCELLED Reason=Nodes_required_for_job_are_DOWN,_DRAINED_or_reserved_for_jobs_in_higher_priority_partitions
+           JobState=CANCELLED Reason=Nodes_required_for_job_are...
        Dependency=(null)
            Requeue=1 Restarts=0 BatchFlag=0 Reboot=0 ExitCode=0:0
            RunTime=00:00:00 TimeLimit=UNLIMITED TimeMin=N/A
@@ -99,7 +99,7 @@ async def test_get_licenses_for_job_no_licenses(
         JobId=2 JobName=sleep
            UserId=root(0) GroupId=root(0) MCS_label=N/A
            Priority=4294901759 Nice=0 Account=root QOS=normal
-           JobState=CANCELLED Reason=Nodes_required_for_job_are_DOWN,_DRAINED_or_reserved_for_jobs_in_higher_priority_partitions
+           JobState=CANCELLED Reason=Nodes_required_for_job_are...
        Dependency=(null)
            Requeue=1 Restarts=0 BatchFlag=0 Reboot=0 ExitCode=0:0
            RunTime=00:00:00 TimeLimit=UNLIMITED TimeMin=N/A
--- a/src/licensemanager2/test/agent/test_slurmctld_prolog.py
+++ b/src/licensemanager2/test/agent/test_slurmctld_prolog.py
@@ -129,7 +129,7 @@ async def test_get_licenses_for_job_no_licenses(
 
 @mark.asyncio
 @mock.patch("asyncio.create_subprocess_shell")
-async def test_get_licenses_for_job_error(
+async def test_get_licenses_for_job_command_error(
     create_process_shell_mock: mock.MagicMock,
 ):
     """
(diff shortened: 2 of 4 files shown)


src/licensemanager2/test/agent/test_slurmctld_prolog.py
```diff
@@ -32,6 +35,114 @@ def slurm_job_id() -> str:
...
+           SuspendTime=None SecsPreSuspend=0 LastSchedEval=2021-07-19T20:22:46
+           Partition=osd-slurmd AllocNode:Sid=juju-fe651b-2:8232
+           ReqNodeList=(null) ExcNodeList=(null)
+           NodeList=(null)
+           NumNodes=1 NumCPUs=8 NumTasks=0 CPUs/Task=1 ReqB:S:C:T=0:0:*:*
+           TRES=cpu=8,billing=8
+           Socks/Node=* NtasksPerN:B:S:C=0:0:*:* CoreSpec=*
+           MinCPUsNode=1 MinMemoryNode=0 MinTmpDiskNode=0
+           Features=(null) DelayBoot=00:00:00
+              OverSubscribe=NO Contiguous=0 Network=(null)
```
jamesbeedy:
```suggestion
              OverSubscribe=NO Contiguous=0 Licenses=(null) Network=(null)
```

src/licensemanager2/test/agent/test_slurmctld_prolog.py
```diff
@@ -32,6 +35,114 @@ def slurm_job_id() -> str:
...
+    assert license_list == ["test_feature.test_product@flexlm:10"]
+
+
+@mark.asyncio
+@mock.patch("asyncio.create_subprocess_shell")
+async def test_get_licenses_for_job_no_licenses(
+    create_process_shell_mock: mock.MagicMock,
+):
+    """
+    Do I return an empty list if the regular expression fails to match any licenses
```
jamesbeedy: Possibly what we are testing here is that the `get_licenses_for_job(job_id)` function returns the empty list if no licenses were specified for the `job_id`.

A rewording might be in order here.

```suggestion
Do I return an empty list if the job doesn't specify any licenses?
```

src/licensemanager2/test/agent/test_slurmctld_prolog.py
jamesbeedy: `scontrol show job=<job_id>` should always return a `Licenses` resource in it's output, even if there were no licenses used for the job. In the case the job doesn't specify licenses then value will be `(null)`.

src/licensemanager2/test/agent/test_slurmctld_prolog.py
```diff
@@ -32,6 +35,114 @@ def slurm_job_id() -> str:
...
+           Features=(null) DelayBoot=00:00:00
+              OverSubscribe=NO Contiguous=0 Licenses=test_feature.test_product@flexlm:10 Network=(null)
+           Command=sleep
+           WorkDir=/home/ubuntu
+           Power=
+           NtasksPerTRES:0
+        """,
+        returncode=0,
+    )
+    license_list = await get_licenses_for_job("some-job")
```
jamesbeedy:
```suggestion
    license_list = await get_licenses_for_job("some-job-id")
```

src/licensemanager2/test/agent/test_slurmctld_prolog.py
```diff
@@ -32,6 +35,114 @@ def slurm_job_id() -> str:
...
+           Features=(null) DelayBoot=00:00:00
+              OverSubscribe=NO Contiguous=0 Network=(null)
+           Command=sleep
+           WorkDir=/home/ubuntu
+           Power=
+           NtasksPerTRES:0
+        """,
+        returncode=0,
+    )
+    license_list = await get_licenses_for_job("some-job")
```
jamesbeedy:
```suggestion
    license_list = await get_licenses_for_job("some-job-id")
```

src/licensemanager2/test/agent/test_slurmctld_prolog.py
```diff
@@ -32,6 +35,114 @@ def slurm_job_id() -> str:
...
+):
+    """
+    Do I raise an ScontrolRetrievalFailure exception if the return value is not 0?
+    """
+    create_process_shell_mock.return_value = MockProcess(
+        "",
+        returncode=1,
+    )
+    with raises(ScontrolRetrievalFailure, match="Could not get SLURM data for job id: some-job"):
+        await get_licenses_for_job("some-job")
```
jamesbeedy:
```suggestion
        await get_licenses_for_job("some-job-id")
```

pyproject.toml
```diff
@@ -106,7 +106,12 @@ mypy = [
     "types-Jinja2",
     "types-jwt",
     "types-click",
-    ]
+    "sqlalchemy-stubs",
+    "httpx",
+    "pytest",
```
zsinx6, line 111: Why not remove the mypy and put everything related to some kind of test, inside a test/lint extra?

src/licensemanager2/test/agent/test_slurmctld_prolog.py
```diff
@@ -32,6 +35,114 @@ def slurm_job_id() -> str:
...
+        return (self.result, None)  # Must be at least a 2-tuple
+
+
+@mark.asyncio
+@mock.patch("asyncio.create_subprocess_shell")
+async def test_get_licenses_for_job_good(
+    create_process_shell_mock: mock.MagicMock,
+):
+    """
+    Do I properly parse licenses out of `scontrol show` commands?
```
jamesbeedy:
```suggestion
    Do I properly parse licenses out of `scontrol show` command?
```

pyproject.toml
dusktreader: I'm not sure the use-cases here. Since I was just trying to fix a mypy but, I used that extra. I've usually worked with packages that just do a standard and dev. I'd like to hear what others think of this suggestion.

jamesbeedy: I think it is important to have distinct extras for each additional entrypoint defined by the project in the case that the respective entrypoints don't require the same deps. For test dependencies where they are likely to all be installed in the same environment I don't have as much of an opinion.

For example, the dependencies of [createjwt.py](https://github.com/omnivector-solutions/license-manager/blob/main/src/licensemanager2/backend/createjwt.py#L7,L10). An end user may want to use this utility from a separate location then where the license-manager-agent is installed. In this case, the user doesn't need all the deps installed to be able to run `lm-create-jwt`.

jamesbeedy: I'm ok with squashing the test deps so long as we keep distinct extras defined for the entrypoints.

dusktreader: For now, then, I will keep the mypy extras defined separately

jamesbeedy: Lets talk about separating them and make a separate pr for that if we decide to do so.

jamesbeedy:
👍

zsinx6 (approved):

jamesbeedy:
👍

PR #36, opened 2021-07-22
get squeue formatted output and parse it

jamesbeedy:
#### What
These changes introduce the squeue_parser() function and associated code including; `_total_time_in_seconds()` and
`return_parsed_squeue_output()` as well as a test that makes assertions on 3 different cases; 1) the case of days in the time string, 2) the case of no days, but hours and 3) the case of only minutes and seconds.

#### Why
Part of the the booking cleanup work, implementing a way to get a list of running jobs on the cluster using squeue.

`Task`: https://omnivector-solutions.monday.com/boards/1204470680/pulses/1456711228

---

#### Peer Review

### DIFF
--- a/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
+++ b/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
@@ -8,5 +8,11 @@ def test_squeue_parser_returns_correct_output_format():
         {"job_id": 2, "run_time_in_seconds": 184981, "state": "RUNNING"},
         {"job_id": 3, "run_time_in_seconds": 17084, "state": "RUNNING"},
     ]
-    squeue_parsed = squeue_parser("1|5:00|RUNNING\n2|2-3:23:01|RUNNING\n3|4:44:44|RUNNING")
+    squeue_parsed = squeue_parser(
+        "\n".join([
+            "1|5:00|RUNNING",
+            "2|2-3:23:01|RUNNING",
+            "3|4:44:44|RUNNING",
+        ])
+    )
     assert squeue_parsed == squeue_parsed_output
--- a/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
+++ b/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
@@ -1,4 +1,13 @@
-from licensemanager2.workload_managers.slurm.cmd_utils import squeue_parser
+from licensemanager2.workload_managers.slurm.cmd_utils import (
+    squeue_parser,
+    SqueueParserUnexpectedInputError,
+)
+
+
+def test_squeue_with_bad_input():
+    """Test that squeue throws the correct exception given incorrect input."""
+    with raises(SqueueParserUnexpectedInputError):
+        squeue_parsed = squeue_parser("bad input")
 
 
 def test_squeue_parser_returns_correct_output_format():
--- a/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
+++ b/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
@@ -2,7 +2,7 @@ from licensemanager2.workload_managers.slurm.cmd_utils import squeue_parser
 
 
 def test_squeue_parser_returns_correct_output_format():
-    """Test that the squeue_parsed() returns correct output."""
+    """Given the squeue formatted output, ensure the `squeue_parsed()` returns the expected values."""
     squeue_parsed_output = [
         {"job_id": 1, "run_time_in_seconds": 300, "state": "RUNNING"},
         {"job_id": 2, "run_time_in_seconds": 184981, "state": "RUNNING"},
--- a/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
+++ b/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
@@ -1,0 +1,12 @@
+from licensemanager2.workload_managers.slurm.cmd_utils import squeue_parser
+
+
+def test_squeue_parser_returns_correct_output_format():
+    """Test that the squeue_parsed() returns correct output."""
+    squeue_parsed_output = [
+        {"job_id": 1, "run_time_in_seconds": 300, "state": "RUNNING"},
+        {"job_id": 2, "run_time_in_seconds": 184981, "state": "RUNNING"},
+        {"job_id": 3, "run_time_in_seconds": 17084, "state": "RUNNING"},
+    ]
+    squeue_parsed = squeue_parser("1|5:00|RUNNING\n2|2-3:23:01|RUNNING\n3|4:44:44|RUNNING")
+    assert squeue_parsed == squeue_parsed_output
--- a/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
+++ b/src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
@@ -1,3 +1,5 @@
+from pytest import raises
+
 from licensemanager2.workload_managers.slurm.cmd_utils import (
     squeue_parser,
     SqueueParserUnexpectedInputError,
@@ -7,7 +9,7 @@ from licensemanager2.workload_managers.slurm.cmd_utils import (
 def test_squeue_with_bad_input():
     """Test that squeue throws the correct exception given incorrect input."""
     with raises(SqueueParserUnexpectedInputError):
-        squeue_parsed = squeue_parser("bad input")
+        _ = squeue_parser("bad input")
 
 
 def test_squeue_parser_returns_correct_output_format():
--- a/src/licensemanager2/workload_managers/slurm/cmd_utils.py
+++ b/src/licensemanager2/workload_managers/slurm/cmd_utils.py
@@ -3,17 +3,19 @@ import asyncio
 import httpx
 import re
 import shlex
+import subprocess
 
 from pydantic import BaseModel, Field
 from licensemanager2.workload_managers.slurm.common import (
     CMD_TIMEOUT,
     SCONTROL_PATH,
     SACCTMGR_PATH,
+    SQUEUE_PATH,
     ENCODING,
     LM2_AGENT_HEADERS,
 )
 from licensemanager2.agent import log as logger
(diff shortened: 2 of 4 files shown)


src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -294,3 +296,61 @@ async def sacctmgr_modify_resource(
...
+
+    try:
+        stdout, _ = subprocess.Popen(
+            [SQUEUE_PATH, "--noheader", "--format='%A|%M|%T'"]
+        )
+    except subprocess.CalledProcessError as e:
+        logger.error(e)
+        raise e
+
+    return stdout.decode().strip()
```
zsinx6: Fix the subprocess usage and also fixes mypy problems
```suggestion
     """Call squeue via Popen and return the output."""

    result = subprocess.run(
        [SQUEUE_PATH + " --noheader --format='%A|%M|%T'"],
        capture_output=True,
        shell=True,
        encoding="utf-8"
    )
    if result.returncode != 0:
        logger.error(result.stderr)
        raise Exception(result.stderr)

    return result.stdout.strip()
```

zsinx6: The  `raise Exception(result.stderr)` should not be necessary just keeped it because was raising a exception. Do you think it makes sense to raise one here?
And the `CalledProcessError` is not the correct one for this case, check [here](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError)

zsinx6 (changes requested):
We should keep the CI always passing, I proposed some fixes to all the mypy problems, take a look

zsinx6:
And what do you think of using this PR to insert the PR Template as well? So we can always remember to update the monday board and also put the link in the PR description together with the standard description text.

src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -294,3 +296,61 @@ async def sacctmgr_modify_resource(
...
+    measure_of_time = time_string.count(":")
+
+    if measure_of_time == 2:
+        hours, minutes, seconds = time_string.split(":")
+    else:
+        minutes, seconds = time_string.split(":")
+
+    return (
+        int(days) * DAY + int(hours) * HOUR + int(minutes) * MINUTE + int(seconds)
+    )
```
zsinx6: Fix mypy and also a more clean way of parsing, since we are already importing `re`
```suggestion
    days = 0
    hours = 0
    splitted_time = [int(value) for value in re.split("-|:", time_string)]
    splitted_time_len = len(splitted_time)

    if splitted_time_len == 3:
        days, hours, minutes, seconds = splitted_time
    elif splitted_time_len == 2:
        hours, minutes, seconds = splitted_time
    else:
        minutes, seconds = splitted_time

    return days * DAY + hours * HOUR + minutes * MINUTE + seconds
```

zsinx6 (approved):

zsinx6:
It is possible to link directly to the `task` in the PR description: https://omnivector-solutions.monday.com/boards/1204470680/pulses/1456711228

src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
```diff
@@ -0,0 +1,12 @@
+from licensemanager2.workload_managers.slurm.cmd_utils import squeue_parser
+
+
+def test_squeue_parser_returns_correct_output_format():
+    """Test that the squeue_parsed() returns correct output."""
```
dusktreader: This docstring is almost a direct duplicate of the method name. Probably not necessary

src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
```diff
@@ -0,0 +1,12 @@
...
+    squeue_parsed_output = [
+        {"job_id": 1, "run_time_in_seconds": 300, "state": "RUNNING"},
+        {"job_id": 2, "run_time_in_seconds": 184981, "state": "RUNNING"},
+        {"job_id": 3, "run_time_in_seconds": 17084, "state": "RUNNING"},
+    ]
+    squeue_parsed = squeue_parser("1|5:00|RUNNING\n2|2-3:23:01|RUNNING\n3|4:44:44|RUNNING")
```
dusktreader: We should probably find a way to break this up into multiple lines so that it reads clearly in the unit tests.

One could either use a modified dedent, or an approach like:
```python
'\n'.join([
    "1|5:00|RUNNING",
    "2|2-3:23:01|RUNNING",
    "3|4:44:44|RUNNING",
])
```

src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
```diff
@@ -0,0 +1,12 @@
+from licensemanager2.workload_managers.slurm.cmd_utils import squeue_parser
+
+
+def test_squeue_parser_returns_correct_output_format():
```
dusktreader, line 15: Though it's unlikely, we should probably provide a unit test that ensures a coherent response to output that doesn't match what is expected.

src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -294,3 +296,58 @@ async def sacctmgr_modify_resource(
         logger.warning(modify_resource_stdout)
         return False
     return True
+
+
+def return_formated_squeue_out() -> str:
+    """Call squeue via Popen and return the output."""
```
dusktreader: This docstring should probably describe the format of the output.

src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -294,3 +296,58 @@ async def sacctmgr_modify_resource(
...
+    )
+    if result.returncode != 0:
+        logger.error(result.stderr)
+        raise Exception(result.stderr)
+
+    return result.stdout.strip()
+
+
+def _total_time_in_seconds(time_string: str) -> int:
+    """Return the sum of the hours minutes and seconds (in seconds)."""
```
dusktreader: This should also mention days if we are calling out hours and minutes explicitly. Alternatively, the docstring could explain that it's converting a specific format to seconds and possibly link to the docs where the format is described

src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -294,3 +296,58 @@ async def sacctmgr_modify_resource(
...
+    MINUTE = 60
+    HOUR = 60 * MINUTE
+    DAY = 24 * HOUR
+
+    days = 0
+    hours = 0
+    splitted_time = [int(value) for value in re.split("-|:", time_string)]
```
dusktreader, line 339: Though unlikely, we should probably have some error handling here.

src/licensemanager2/workload_managers/slurm/cmd_utils.py
jamesbeedy:
```suggestion
    """
    Call squeue via Popen and return the formatted output.
    
    Return the squeue output in the form "<job_id>|<run_time>|<state>".
    """
```

src/licensemanager2/workload_managers/slurm/cmd_utils.py
jamesbeedy:
```suggestion
    """
    Return the runtime in seconds for a job.
    
    This function takes a slurm time string ("<days>-<hours>:<minutes>:<seconds>") as input, parses
    and converts each of the units in the time string to seconds and returns a computed value, the sum of the days,
    hours, minutes and seconds (in seconds).
    """
```

src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
jamesbeedy:
```suggestion
    squeue_parsed = squeue_parser(
        "\n".join([
            "1|5:00|RUNNING",
            "2|2-3:23:01|RUNNING",
            "3|4:44:44|RUNNING",
        ])
    )
```

src/licensemanager2/test/workload_managers/slurm/cmd_test_utils.py
jamesbeedy:
```suggestion
    """Given the squeue formatted output, ensure the `squeue_parsed()` returns the expected values."""
```

src/licensemanager2/workload_managers/slurm/cmd_utils.py
jamesbeedy: I feel like this is less important now that we raise an exception [here](https://github.com/omnivector-solutions/license-manager/pull/36/commits/720e6d2159435b49c22f574543b1b81884c266f7), thoughts?

dusktreader: agreed

src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -3,23 +3,30 @@
...
+from typing import Dict, List, Optional, Union
 from licensemanager2.agent.settings import (
     SETTINGS,
     PRODUCT_FEATURE_RX,
 )
 
 
+class SqueueParserUnexpectedInputError(Exception):
+    """Unexpected squeue output."""
+    pass
```
zsinx6: This can be removed
```suggestion
```

src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -294,3 +301,77 @@ async def sacctmgr_modify_resource(
...
+
+    squeue_parsed_output = list()
+
+    def parse_squeue_line():
+        """Parse a line from squeue formatted output."""
+        try:
+            job_id, run_time, state = line.split("|")
+        except SqueueParserUnexpectedInputError as e:
+            logger.error(e)
+            raise(e)
```
zsinx6:
```suggestion
        except SqueueParserUnexpectedInputError as e:
            logger.error(e)
            raise e
```

dusktreader (approved):

zsinx6 (approved):

PR #38, opened 2021-08-03
Fix squeue parser

zsinx6:
#### What
Fixed tests and the actual code for the squeue parser.

#### Why
The test filename was wrong and the test was not actually running the tests.

`Task`: https://omnivector-solutions.monday.com/boards/1204470680/pulses/1541868998


heitorPB (approved):
:+1:

PR #40, opened 2021-08-04
Ensure packages needed for agent are not optional

zsinx6:
#### What
Make sure all packages that are used by agent are not optional.

#### Why
To be able to run the agent.


jamesbeedy:
👍

PR #42, opened 2021-08-05
Create functions to delete booking using grace_time

zsinx6:
#### What
Delete booking in reconcile if the running time is greater than the grace_time.

#### Why
We need to remove the bookings after the job is over its grace_time.

`Task`: https://omnivector-solutions.monday.com/boards/1204470680/pulses/1456711228


src/licensemanager2/agent/reconciliation.py
```diff
@@ -2,18 +2,105 @@
...
+class FailedToRemoveBookedViaGraceTime(Exception):
+    """Unable to delete booking row."""
+
+
+async def remove_booked_for_job_id(job_id: str):
+    """
+    Send DELETE to /api/v1/booking/book/{job_id}.
+    """
+    response = await async_client().delete("/api/v1/booking/book{job_id}")
+    if response.status_code != 200:
```
dusktreader: It's probably best to use constants for error codes. 200 is pretty trivial, but when it comes to handling stuff in the 400 range I like to have the human-readable constants:

httpx.codes.OK (200)
httpx.codes.BAD_REQUEST (400)
httpx.codes.FORBIDDEN (403)
etc...

src/licensemanager2/agent/reconciliation.py
```diff
@@ -2,18 +2,105 @@
...
+        raise FailedToRemoveBookedViaGraceTime()
+
+
+async def get_all_grace_times() -> Dict[int, int]:
+    """
+    Send GET to /api/v1/config/all.
+    """
+    response = await async_client().get("/api/v1/config/all")
+    configs = response.json()
+    grace_times = {}
```
dusktreader: We could build this with a dict comprehension:

```python
grace_times = {c["id"]: c["grace_time"] for c in configs}
```

src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -307,15 +307,15 @@ async def sacctmgr_modify_resource(
...
+def return_formatted_squeue_out() -> str:
     """
     Call squeue via Popen and return the formatted output.
 
     Return the squeue output in the form "<job_id>|<run_time>|<state>".
     """
 
     result = subprocess.run(
-        [SQUEUE_PATH, "--noheader", "--format='%A|%M|%T'"],
+        f"{SQUEUE_PATH} --noheader --format='%A|%M|%T'",
```
jamesbeedy: can we use `shlex.quote()` here?

src/licensemanager2/agent/reconciliation.py
```diff
@@ -2,18 +2,105 @@
...
+    if not formatted_squeue_output:
+        return
+    squeue_running_jobs = squeue_parser(formatted_squeue_output)
+    grace_times = await get_all_grace_times()
+    get_booked_call = []
+    job_id_greatest_grace_time = {}
+    for job in squeue_running_jobs:
+        job_id = job["job_id"]
+        get_booked_call.append(get_booked_for_job_id(job_id))
+        job_id_greatest_grace_time[job_id] = 0
```
dusktreader: I don't think we need to pre-seed this with 0s. Instead, in the loop below (`for job_id in job_id_greatest_grace_time.keys()`) we can instead just loop over job in sque_running_jobs again

src/licensemanager2/agent/reconciliation.py
```diff
@@ -2,18 +2,105 @@
...
+    booking_rows_for_running_jobs = asyncio.gather(*get_booked_call)
+
+    # get the greatest grace_time for each job
+    for job_id in job_id_greatest_grace_time.keys():
+        job_id_greatest_grace_time[job_id] = get_greatest_grace_time(job_id,
+                                                                     grace_times,
+                                                                     booking_rows_for_running_jobs)
+
+    # for each job, if the running_time is greater than the greatest grace_time, delete the booking for it
+    for job in squeue_running_jobs:
```
dusktreader: This loop and the previous one could be collapsed together, couldn't they?

src/licensemanager2/test/agent/test_reconciliation.py
```diff
@@ -0,0 +1,150 @@
...
+        },
+    ]
+
+
+def test_get_greatest_grace_time(booking_rows_json):
+    """
+    Test if the function really returns the greatest value for the grace_time.
+    """
+    grace_times = {
+        1: 100,
```
dusktreader: I think we should have more values here to test edge-cases more thoroughly

src/licensemanager2/test/agent/test_reconciliation.py
```diff
@@ -0,0 +1,150 @@
...
+    booking_rows_json,
+):
+    """
+    Test for when the running time is greater than the grace_time, then delete the booking.
+    """
+    asyncio_mock.gather.return_value = booking_rows_json
+    return_formatted_squeue_out_mock.return_value = "1|5:00|RUNNING"
+    get_all_grace_times_mock.return_value = {1: 100, 2: 30}
+    await clean_booked_grace_time()
+    remove_booked_for_job_id_mock.assert_awaited_once()
```
dusktreader: should this check the parameters passed?

src/licensemanager2/workload_managers/slurm/cmd_utils.py
```diff
@@ -354,10 +354,13 @@ def _total_time_in_seconds(time_string: str) -> int:
     return days * DAY + hours * HOUR + minutes * MINUTE + seconds
 
 
-def squeue_parser(squeue_formatted_output) -> List[Dict]:
+def squeue_parser(squeue_formatted_output) -> List:
```
dusktreader: I wonder if it would be useful to create a pydantic model for what job items contain. I think it might help the code be more self-documenting and the access/usage in the gracetime cleanup methods might be a bit smoother.

dusktreader (changes requested):

src/licensemanager2/workload_managers/slurm/cmd_utils.py
zsinx6: Yes, but since the entire string is something we provide ourselves, does it make sense to use?

src/licensemanager2/test/agent/test_reconciliation.py
zsinx6: Do you have any suggestion?

src/licensemanager2/workload_managers/slurm/cmd_utils.py
zsinx6: I don't think it will help much, the result is already very simple, just a dict with 3 keys

src/licensemanager2/workload_managers/slurm/cmd_utils.py
jamesbeedy: See [here](https://docs.python.org/3/library/shlex.html#shlex.quote), templating the variable into the command, even in our use case exposes the security hole.

jamesbeedy: We have tried to make it a convention to use this throughout license-manager.

jamesbeedy:
👍

dusktreader (approved):

jamesbeedy:
👍

Issue #41, opened 2021-08-04
split the project into two


jamesbeedy:
We need to split the license-manager project into two separate projects. The license-manager has outgrown the confines of living in a single python project.

Split the license-manager up into two separate projects; `license-manager-backend`, `license-manager-agent`.

PR #43, opened 2021-08-06
Separated backend and agent code into separate sub-projects

dusktreader:
#### What
Divided agent and backend code into separate sub-projects:
* each is independently managed with its own pypproject.toml
* no shared code
* removed some derelict code and small refactors
* Makefiles updated with new convenience commands
* root Makefile triggers sub-projects' Makefiles

#### Why
The two code bases shared little code. However, having a single set of dependencies and configuration was difficult to maintain. It needed to be easier to work on one side or the other independently, so it made sense to create two independent sup-projects

`Task`: https://omnivector-solutions.monday.com/boards/1204470680/pulses/1547838946

---

#### Peer Review


Makefile
```diff
@@ -1,30 +1,19 @@
...
-            && zip -9 -q ../../../../$@ -r . \
-                && cd ../../../../
-	rm -rf _tmp_venvi
+	rm -rf $@
+	cd backend && python3 -m venv _tmp_venv
+	cd backend && poetry config virtualenvs.create false --local
+	cd backend && source _tmp_venv/bin/activate && poetry install --no-dev
+	cd backend && cd _tmp_venv/lib/python*.*/site-packages && zip -qr9 ${ROOT_DIR}/$@ .
+	rm -rf backend/_tmp_venv
+	rm backend/poetry.toml
```
zsinx6: What do you think about using pip instead of poetry here? I think the Makefile should works always, and be agnostic of poetry.
And also, I think it makes sense to have a Makefile inside the `backend` folder, this way you don't need to have to `cd` everytime, and also makes this root Makefile more readable to calling other scripts (e.g. format, lint, test...)

```bash
function.zip:
        $(MAKE) -C backend function.zip
```

backend/.gitignore
```diff
@@ -0,0 +1,4 @@
+.coverage
+sqlite.db
```
zsinx6: Since this is at both `agent` and `backend`, it should be moved to the root gitignore

agent/app/api/__init__.py
```diff
@@ -1,13 +1,10 @@
-"""
-Endpoints for the agent
-"""
 from fastapi import APIRouter
 
-from licensemanager2.agent.booking_proxy import booking_proxy_router
-from licensemanager2.agent.license_proxy import license_proxy_router
+from app.api.booking_proxy import router as booking_proxy_router
+from app.api.license_proxy import router as license_proxy_router
```
zsinx6: Why app?

Makefile
dusktreader: I think it's a great idea to have makefiles in the subdirectories to simplify the united makefile *a lot*.

However, we are using poetry to manage dependencies, so I think we will need to use it for the build as well. I think it's a baseline dependency for your dev environment and for our CI/CD builds, so I don't think we're making too many assumptions that it will be available.

backend/.gitignore
dusktreader: Yeah, that's a good point. I was working on them independently, so I added some things to their .gitignore files that should go in the root.

agent/app/api/__init__.py
dusktreader: There's a few reasons for this.

First, you can change your package name without affecting your imports. For example, if we forked and wanted a license-manager v3, we would not have to change all the imports.

Second, it allows you to have absolute imports with a _short_ name that's not going to collide with a dependency installed through pypi.

Finally, it's a re-usable pattern that you can include in any project that is immediately familiar.

I prefer to use `app` as the root package name when I'm building out stand-alone applications. In the case of building a package/library then I would use the name of the package as the root.

zsinx6: Does it make sense to change the package name in the first place?
How can the absolute import conflict with something installed via pypi?

About the last point, I didn't like it, it makes harder to know where did that code comes from, when reading only a part of the file

dusktreader: Package names can often change. It's nice if each individual component of an app doesn't have to know what the containing package is named. It's also nice to have the root name be short so your project name can still be long. For instance, I may want my application project to be named "license-manager-agent". It's nice to import from "app" instead of "license_manager_agent".

Collisions with names on pypi don't happen often, but they can. It's nice for it to be very obvious that you are importing something from your local app definitions as opposed to a package available on pypi.

I understand the last point. However, I don't often find myself looking at a file where I've forgotten the project I'm working and it's usually easy to see the path to know what I'm looking at.

I think this is a good thing to discuss. If we do not go with 'app', I would lean towards using 'backend' and 'agent' as the root package names for both.

jamesbeedy: I like removing the ambiguity here, how about being even slightly more verbose, `app` -> `license_manager_backend`?

heitorPB: `app` is very ambiguous. `lm_backend` could be a good compromise between short and descriptive.

jamesbeedy: So the folder structure for the backend would resemble:
```bash
license-manager/
license-manager/license-manager-backend/
license-manager/license-manager-backend/pyproject.toml
license-manager/license-manager-backend/src/
license-manager/license-manager-backend/src/lm_backend/
license-manager/license-manager-backend/src/lm_backend/main.py
```

?

zsinx6: @jamesbeedy will be something like this:

`license-manager/backend/pyproject.toml`
`license-manager/backend/license_manager_backend/main.py `

jamesbeedy (approved):
👍

agent/app/api/__init__.py
dusktreader: I've adjusted it so that app is now lm_agent and lm_backend respectively

agent/.python-version
```diff
@@ -0,0 +1 @@
+3.8.10
```
zsinx6: What is this for?

dusktreader: This is a file that pyenv creates. Missed it in my .gitignore

jamesbeedy (approved):
👍

zsinx6:
This file can be deleted: https://github.com/omnivector-solutions/license-manager/blob/tucker/separate-backend-and-agent/pyproject.toml

Edit: Pushed this change

dusktreader:
> This file can be deleted: https://github.com/omnivector-solutions/license-manager/blob/tucker/separate-backend-and-agent/pyproject.toml
> 
> Edit: Pushed this change

Thanks. Not sure how I missed that one, lol

jamesbeedy:
Fixes: #41

PR #46, opened 2021-08-13
don't start the agent if the backend version doesn't match

### LINKED ISSUE #45: Enforce consistency between the agent and backend
Currently the license-manager-agent has no way of knowing if the license-manager-backend is compatible or not. Add some verification and prevent the license-manager-agent from starting if the version doesn't match the backend.

jamesbeedy:
#### What
Create a /version backend endpoint that exposes the version
of the license-manager-backend. Additionally add a startup
check in the license-manager-agent to ensure that it is
running the same code as the backend. Additionally, add a /debug
endpoint to the agent that exposes the version information of the
agent and the backend, to allow for simplified debugging.

#### Why
We need a way to guarantee that the license-manager-agent is talking to a compatible license-manager-backend and want to simplify the process of being able to introspect what versions are running/configured.

Fixes: #45

### DIFF
--- a/agent/lm_agent/main.py
+++ b/agent/lm_agent/main.py
@@ -4,23 +4,22 @@ License-manager agent, command line entrypoint
 Run with e.g. `uvicorn lm_agent.main:app`
 """
 import logging
-import pkg_resources
 import typing
 from itertools import cycle
 
+import pkg_resources
 from fastapi import FastAPI, HTTPException
 from fastapi_utils.tasks import repeat_every
 
 from lm_agent.api import api_v1
 from lm_agent.backend_utils import (
-    LicenseManagerBackendVersionError,
     LicenseManagerBackendConnectionError,
+    LicenseManagerBackendVersionError,
 )
 from lm_agent.config import settings
 from lm_agent.forward import async_client
 from lm_agent.logs import init_logging, logger
 from lm_agent.reconciliation import reconcile
-
 
 app = FastAPI()
 # app.add_middleware(
@@ -71,7 +70,7 @@ def backend_version_check():
 
     # Check the version of the backend matches the version of the agent.
     backend_version = resp.json()["version"]
-    agent_version = pkg_resources.get_distribution('license-manager-agent').version
+    agent_version = pkg_resources.get_distribution("license-manager-agent").version
     if backend_version != agent_version:
         logger.error(f"license-manager-backend incompatible version: {backend_version}.")
         raise LicenseManagerBackendVersionError()
--- a/agent/lm_agent/main.py
+++ b/agent/lm_agent/main.py
@@ -87,7 +87,7 @@ async def backend_version_check():
 
     # Check the version of the backend matches the version of the agent.
     BACKEND_VERSION = resp.json()["version"]
-    if BACKEND_VERSION != AGENT_VERSION:
+    if BACKEND_VERSION.split(".")[0] != AGENT_VERSION.split(".")[0]:
         logger.error(f"license-manager-backend incompatible version: {BACKEND_VERSION}.")
         raise LicenseManagerBackendVersionError()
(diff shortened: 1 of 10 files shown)


agent/lm_agent/main.py
```diff
@@ -52,6 +58,25 @@ async def reconcile_endpoint():
...
+    resp = await async_client().get("/version")
+    # Check that we have a valid response.
+    if resp.status_code != status.HTTP_200_OK:
+        logger.error("license-manager-backend version could not be obtained.")
+        raise LicenseManagerBackendConnectionError()
+
+    # Check the version of the backend matches the version of the agent.
+    backend_version = resp.json()["version"]
+    agent_version = pkg_resources.get_distribution("license-manager-agent").version
+    if backend_version != agent_version:
```
zsinx6: Will we need to enforce always the exactly same version?

backend/pyproject.toml
```diff
@@ -35,6 +35,7 @@ black = "^21.7b0"
 alembic = "^1.6.5"
 mypy = "^0.910"
 sqlalchemy-stubs = "^0.4"
+types-setuptools = "^57.0.2"
```
zsinx6: I think this makes more sense to go here: https://github.com/omnivector-solutions/license-manager/blob/main/backend/mypy.ini
Don't think it will add much in this case, just a new dependency

agent/lm_agent/main.py
```diff
@@ -52,6 +58,25 @@ async def reconcile_endpoint():
     await reconcile()
 
 
+@app.on_event("startup")
+def backend_version_check():
```
zsinx6:
```suggestion
async def backend_version_check():
```

zsinx6: We also need a new test for this new function

agent/lm_agent/main.py
jamesbeedy: Possibly we should only enforce the same major version, and not the exact version?

zsinx6: I think the agent should work with backend.version >= agent.version, i.e. if agent is 1.2.3 and backend is 1.2.4 or 1.3.1 or 1.5.0 should also work, just ensure if the major is the same and the backend version is greater. What do you think?

zsinx6: I think this should help our workflow, not the opposite. The backend is independent of the license server, it is just a API, then is it way more stable. The agent on the other hand is more difficult to develop and depends on the license server, and updates there are more frequent, and as the versions of the agent updates, we want to still use the same version for the backend during the entire development

jamesbeedy: Take the case where the agent is 1.x and the backend is 2.x. We know the agent will not work with the backend and therefore should not attempt to start it. The inverse (where the agent is 2.x and the backend is 1.x) is also a case where we do not want to start the agent. I think a sane major version check might do the job here. Thoughts?

zsinx6: I agree, but this way we will always move the major version of the backend together with the agent? Always?

backend/mypy.ini
```diff
@@ -19,3 +19,6 @@ ignore_missing_imports = True
 
 [mypy-typer.*]
 ignore_missing_imports = True
+
+[mypy-pkgresources.*]
```
zsinx6:
```suggestion
[mypy-pkg_resources.*]
```

agent/lm_agent/main.py
```diff
@@ -36,6 +46,17 @@ async def root():
     return dict(message="OK")
 
 
+@app.get("/debug")
+async def debug():
```
zsinx6: Since we have the /version in the backend already, what is the point of replicate the information in the agent?
What do you think about /version in the agent as well, but only returns the agent version?

jamesbeedy: I don't like having them in two separate places because it creates more user interactions to get to the desired information-  that we shouldn't have to reach too far to understand about the application.

PR #48, opened 2021-08-13
add a way to ensure that the agent is compatible with the backend

### LINKED ISSUE #45: Enforce consistency between the agent and backend

jamesbeedy:
#### What
Add an endpoint in the license-manager-backend that exposes
the version. Additionally add a startup check in the agent that
checks for consist major versions between the configured backend
and the running agent.

#### Why
Need a way to provide a guarantee that the running agent will work with the backend before executing agent code.

Fixes: #45 

`Task`: https://omnivector-solutions.monday.com/boards/1204470680/pulses/1575098866

### DIFF
--- a/agent/lm_agent/backend_utils.py
+++ b/agent/lm_agent/backend_utils.py
@@ -3,6 +3,7 @@ Provide utilities that communicate with the backend.
 """
 from typing import List
 
+from fastapi import status
 from httpx import ConnectError
 from pydantic import BaseModel, ValidationError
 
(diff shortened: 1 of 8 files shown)


zsinx6:
Fixes #45. The  #46 is a PR

zsinx6 (approved):

PR #50, opened 2021-08-17
Parser now returns lead_host, user_name and booked

zsinx6:
#### What
Flexlm parser now returns the lead_host, user_name and quantity booked for each use of the licenses.

#### Why
These information are needed by the in use cleanup.

`Task`: https://omnivector-solutions.monday.com/boards/1204470680/pulses/1579578433

### DIFF
--- a/agent/lm_agent/parsing/flexlm.py
+++ b/agent/lm_agent/parsing/flexlm.py
@@ -48,6 +48,7 @@ def parse(s: str) -> dict:
     """
     Parse lines of the license output with regular expressions
     """
+    parsed_data: dict = {"uses": [], "total": None}
     for line in s.splitlines():
         parsed = RX.match(line)
         if parsed is None:
@@ -55,10 +56,18 @@ def parse(s: str) -> dict:
 
         if parsed.group("total"):
             d = parsed.groupdict()
-            return {
+            parsed_data["total"] = {
                 "total": int(d["total"]),
                 "used": int(d["used"]),
                 "feature": d["feature"],
             }
-
-    return {}
+        if parsed.group("user"):
+            d = parsed.groupdict()
+            parsed_data["uses"].append(
+                {
+                    "user_name": d["user"],
+                    "lead_host": d["user_host"],
+                    "booked": int(d["tokens"]),
+                }
+            )
+    return parsed_data
--- a/agent/tests/conftest.py
+++ b/agent/tests/conftest.py
@@ -2,9 +2,9 @@
 Configuration of pytest for agent tests
 """
 from pathlib import Path
+from textwrap import dedent
 from unittest.mock import patch
 
-import pkg_resources
 import respx
 from httpx import ASGITransport, AsyncClient
 from pytest import fixture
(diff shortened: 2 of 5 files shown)


agent/lm_agent/parsing/flexlm.py
```diff
@@ -48,17 +48,26 @@ def parse(s: str) -> dict:
...
             d = parsed.groupdict()
-            return {
+            parsed_data["total"] = {
                 "total": int(d["total"]),
                 "used": int(d["used"]),
                 "feature": d["feature"],
             }
-
-    return {}
+        if parsed.group("user"):
```
dusktreader, line 64: I think it might make sense to do the next line before this:
```python
d = parsed.groupdict()
if d["user"]:
    ...
```

This is especially relevant because we are pulling "user" out of the groupdict later anyway

agent/lm_agent/parsing/flexlm.py
```diff
@@ -48,17 +48,26 @@ def parse(s: str) -> dict:
...
+            d = parsed.groupdict()
+            parsed_data["uses"].append(
+                {
+                    "user_name": d["user"],
+                    "lead_host": d["user_host"],
+                    "booked": int(d["tokens"]),
```
dusktreader, line 70: should we use a default of 0 here for tokens?

agent/tests/conftest.py
```diff
@@ -51,3 +51,43 @@ def respx_mock():
...
+
+
+@fixture
+def lm_output_bad():
+    """
+    Some unparseable lmstat output
+    """
+    return dedent(
+        """\
+    lmstat - Copyright (c) 1989-2004 by Macrovision Corporation. All rights reserved.
```
dusktreader, line 63: you can make this behave in a nicer way by using `strip()` after the dedent:
```python
return dedent(
    """
    lmstat...
    Flexible...
    """
).strip()
```

my `snick` package does this

jamesbeedy (approved):
👍

agent/tests/conftest.py
```diff
@@ -51,3 +51,43 @@ def respx_mock():
...
+        lmstat - Copyright (c) 1989-2004 by Macrovision Corporation. All rights reserved.
+        ...
+
+        Users of TESTFEATURE:  (Total of 1000 licenses issued;  Total of 93 licenses in use)
+
+        ...
+
+
+        """
+        "           jbemfv myserver.example.com /dev/tty (v62.2) (myserver.example.com/24200 12507), "
```
dusktreader, line 87: Can we include this in the triple-quote string? It's a little odd looking with the switch to implicit string concatenation in the middle

dusktreader (changes requested):

agent/lm_agent/parsing/flexlm.py
zsinx6: Will solve this in another PR

agent/lm_agent/parsing/flexlm.py
zsinx6: No, the list must be empty

PR #51, opened 2021-08-18
In use cleanup

zsinx6:
#### What
Implement in use cleanup.

#### Why
We need to clean the in use licenses from the booking table.

`Task`: https://omnivector-solutions.monday.com/boards/1204470680/pulses/1456713032

### DIFF
--- a/agent/lm_agent/main.py
+++ b/agent/lm_agent/main.py
@@ -4,9 +4,10 @@ License-manager agent, command line entrypoint
 Run with e.g. `uvicorn lm_agent.main:app`
 """
 import logging
+from typing import Any
 
-import sentry_sdk
 import pkg_resources
+import sentry_sdk
 from fastapi import FastAPI
 from fastapi_utils.tasks import repeat_every
 from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
@@ -20,7 +21,7 @@ from lm_agent.reconciliation import reconcile
 AGENT_VERSION = pkg_resources.get_distribution("license-manager-agent").version
 
 
-app = FastAPI()
+app: Any = FastAPI()
 # app.add_middleware(
 #     CORSMiddleware,
 #     allow_origin_regex=settings.ALLOW_ORIGINS_REGEX,
--- a/agent/lm_agent/main.py
+++ b/agent/lm_agent/main.py
@@ -32,12 +32,6 @@ app: Any = FastAPI()
 # app.add_middleware(TrustedHostMiddleware)
 # app.add_middleware(ProxyHeadersMiddleware)
 # app.add_middleware(RateLimitMiddleware)
-if settings.SENTRY_DSN:
-    sentry_sdk.init(
-        dsn=settings.SENTRY_DSN,
-        traces_sample_rate=1.0,
-    )
-    app = SentryAsgiMiddleware(app)
 
 
 @app.get("/")
@@ -112,3 +106,10 @@ async def collect_stats():
 
 
 app.include_router(api_v1, prefix="/api/v1")
+
+if settings.SENTRY_DSN:
+    sentry_sdk.init(
+        dsn=settings.SENTRY_DSN,
+        traces_sample_rate=1.0,
+    )
+    app = SentryAsgiMiddleware(app)
--- a/agent/lm_agent/main.py
+++ b/agent/lm_agent/main.py
@@ -4,7 +4,6 @@ License-manager agent, command line entrypoint
 Run with e.g. `uvicorn lm_agent.main:app`
 """
 import logging
-from typing import Any
 
 import pkg_resources
 import sentry_sdk
@@ -21,17 +20,13 @@ from lm_agent.reconciliation import reconcile
 AGENT_VERSION = pkg_resources.get_distribution("license-manager-agent").version
 
 
-app: Any = FastAPI()
-# app.add_middleware(
-#     CORSMiddleware,
-#     allow_origin_regex=settings.ALLOW_ORIGINS_REGEX,
-#     allow_credentials=True,
-#     allow_methods=["*"],
-#     allow_headers=["*"],
-# )
-# app.add_middleware(TrustedHostMiddleware)
-# app.add_middleware(ProxyHeadersMiddleware)
-# app.add_middleware(RateLimitMiddleware)
+app = FastAPI()
+if settings.SENTRY_DSN:
+    sentry_sdk.init(
+        dsn=settings.SENTRY_DSN,
+        traces_sample_rate=1.0,
+    )
+    app.add_middleware(SentryAsgiMiddleware)
 
 
 @app.get("/")
@@ -106,10 +101,3 @@ async def collect_stats():
 
 
 app.include_router(api_v1, prefix="/api/v1")
-
-if settings.SENTRY_DSN:
-    sentry_sdk.init(
-        dsn=settings.SENTRY_DSN,
-        traces_sample_rate=1.0,
-    )
-    app = SentryAsgiMiddleware(app)
--- a/agent/lm_agent/main.py
+++ b/agent/lm_agent/main.py
@@ -5,9 +5,11 @@ Run with e.g. `uvicorn lm_agent.main:app`
 """
 import logging
 
+import sentry_sdk
 import pkg_resources
 from fastapi import FastAPI
 from fastapi_utils.tasks import repeat_every
+from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 
 from lm_agent.api import api_v1
 from lm_agent.backend_utils import LicenseManagerBackendVersionError, get_license_manager_backend_version
@@ -29,6 +31,12 @@ app = FastAPI()
 # app.add_middleware(TrustedHostMiddleware)
 # app.add_middleware(ProxyHeadersMiddleware)
 # app.add_middleware(RateLimitMiddleware)
+if settings.SENTRY_DSN:
+    sentry_sdk.init(
+        dsn=settings.SENTRY_DSN,
+        traces_sample_rate=1.0,
+    )
+    app = SentryAsgiMiddleware(app)
 
 
 @app.get("/")
--- a/agent/lm_agent/reconciliation.py
+++ b/agent/lm_agent/reconciliation.py
@@ -2,7 +2,7 @@
 """
 Reconciliation functionality live here.
 """
-import asyncio  # noqa
+import asyncio
 from typing import Dict, List
 
 from fastapi import HTTPException, status
@@ -86,10 +86,7 @@ async def clean_booked_grace_time():
         job_id = job["job_id"]
         get_booked_call.append(get_booked_for_job_id(job_id))
 
-    booking_rows_for_running_jobs = []
-    for coroutine in get_booked_call:
-        booking_rows_for_running_jobs.append(await coroutine)
-    # booking_rows_for_running_jobs = await asyncio.gather(*get_booked_call)
+    booking_rows_for_running_jobs = await asyncio.gather(*get_booked_call, return_exceptions=True)
 
     # get the greatest grace_time for each job
     for job in squeue_running_jobs:
--- a/agent/lm_agent/reconciliation.py
+++ b/agent/lm_agent/reconciliation.py
@@ -16,18 +16,14 @@ from lm_agent.workload_managers.slurm.cmd_utils import return_formatted_squeue_o
 RECONCILE_URL_PATH = "/api/v1/license/reconcile"
 
 
-class FailedToRemoveBookedViaGraceTime(Exception):
-    """Unable to delete booking row."""
-
-
 async def remove_booked_for_job_id(job_id: str):
     """
     Send DELETE to /api/v1/booking/book/{job_id}.
     """
-    response = await async_client().delete(f"/api/v1/booking/book{job_id}")
+    response = await async_client().delete(f"/api/v1/booking/book/{job_id}")
+    logger.debug(f"response from delete: {response.__dict__}")
     if response.status_code != status.HTTP_200_OK:
         logger.error(f"{job_id} could not be deleted.")
-        raise FailedToRemoveBookedViaGraceTime()
 
 
 async def get_all_grace_times() -> Dict[int, int]:
--- a/agent/lm_agent/reconciliation.py
+++ b/agent/lm_agent/reconciliation.py
@@ -21,9 +21,9 @@ async def remove_booked_for_job_id(job_id: str):
     Send DELETE to /api/v1/booking/book/{job_id}.
     """
     response = await async_client().delete(f"/api/v1/booking/book/{job_id}")
-    logger.debug(f"response from delete: {response.__dict__}")
     if response.status_code != status.HTTP_200_OK:
         logger.error(f"{job_id} could not be deleted.")
+        logger.debug(f"response from delete: {response.__dict__}")
 
 
 async def get_all_grace_times() -> Dict[int, int]:
@@ -41,8 +41,7 @@ async def get_booked_for_job_id(job_id: str) -> Dict:
     Return the booking row for the given job_id.
     """
     response = await async_client().get(f"/api/v1/booking/job/{job_id}")
-    book_row = response.json()
-    return book_row
+    return response.json()
 
 
 def get_greatest_grace_time(job_id: str, grace_times: Dict[int, int], booking_rows: List) -> int:
@@ -88,7 +87,7 @@ async def clean_booked_grace_time():
         job_id = job["job_id"]
         greatest_grace_time = get_greatest_grace_time(job_id, grace_times, booking_rows_for_running_jobs)
         # if the running_time is greater than the greatest grace_time, delete the booking for it
-        if job["run_time_in_seconds"] > greatest_grace_time and greatest_grace_time >= 0:
+        if job["run_time_in_seconds"] > greatest_grace_time and greatest_grace_time != -1:
             await remove_booked_for_job_id(job_id)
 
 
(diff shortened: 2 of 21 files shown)


agent/lm_agent/main.py
```diff
@@ -18,7 +21,7 @@
 AGENT_VERSION = pkg_resources.get_distribution("license-manager-agent").version
 
 
-app = FastAPI()
+app: Any = FastAPI()
```
dusktreader: Shouldn't the type for `app` be `FastAPI`?

agent/lm_agent/main.py
```diff
@@ -103,3 +106,10 @@ async def collect_stats():
 
 
 app.include_router(api_v1, prefix="/api/v1")
+
+if settings.SENTRY_DSN:
+    sentry_sdk.init(
+        dsn=settings.SENTRY_DSN,
+        traces_sample_rate=1.0,
+    )
+    app = SentryAsgiMiddleware(app)
```
dusktreader: I think the idiomatic FastAPI way to do this is:

```python
app.add_middleware(SentryAsgiMiddleware)
```

see: https://fastapi.tiangolo.com/advanced/middleware/#adding-asgi-middlewares

agent/lm_agent/reconciliation.py
```diff
@@ -16,18 +16,14 @@
...
-    """Unable to delete booking row."""
-
-
 async def remove_booked_for_job_id(job_id: str):
     """
     Send DELETE to /api/v1/booking/book/{job_id}.
     """
-    response = await async_client().delete("/api/v1/booking/book{job_id}")
+    response = await async_client().delete(f"/api/v1/booking/book/{job_id}")
+    logger.debug(f"response from delete: {response.__dict__}")
```
dusktreader: do we want to leave this in, or was this only for debugging?

agent/lm_agent/reconciliation.py
```diff
@@ -53,38 +49,51 @@ def get_greatest_grace_time(job_id: str, grace_times: Dict[int, int], booking_ro
     """
     Find the greatest grace_time for the given job_id.
     """
-    greatest_grace_time = 0
+    greatest_grace_time = -1
```
dusktreader, line 51: Should an exception be raised if, after traversing the booking rows, greatest_grase_time is still -1?

agent/lm_agent/reconciliation.py
```diff
@@ -53,38 +49,51 @@ def get_greatest_grace_time(job_id: str, grace_times: Dict[int, int], booking_ro
...
+        if not book:
             continue
-        config_id_for_grace_times = book["config_id"]
-        if grace_times[config_id_for_grace_times] > greatest_grace_time:
-            greatest_grace_time = grace_times[config_id_for_grace_times]
+        for inner_book in book:
+            if str(inner_book["job_id"]) != str(job_id):
+                continue
+            config_id_for_grace_times = inner_book["config_id"]
+            if grace_times[config_id_for_grace_times] > greatest_grace_time:
```
dusktreader: you could replace these two lines with:
```python
greatest_grace_time = max(greatest_gracetime, grace_times[config_id_for_grace_times])
```
I think it might read nicely this way

agent/lm_agent/reconciliation.py
```diff
@@ -53,38 +49,51 @@ def get_greatest_grace_time(job_id: str, grace_times: Dict[int, int], booking_ro
...
+                greatest_grace_time = grace_times[config_id_for_grace_times]
     return greatest_grace_time
 
 
+def get_running_jobs(squeue_result: List) -> List:
+    squeue_running_jobs = []
+    for job in squeue_result:
```
dusktreader: I think this logic could be reduced with a list comprehension (or filter() ):
```python
return [j for j in squeue_result if j["state"] == "RUNNING"]
```

agent/lm_agent/reconciliation.py
```diff
@@ -53,38 +49,51 @@ def get_greatest_grace_time(job_id: str, grace_times: Dict[int, int], booking_ro
...
+    squeue_running_jobs = get_running_jobs(squeue_result)
+
     grace_times = await get_all_grace_times()
     get_booked_call = []
     for job in squeue_running_jobs:
         job_id = job["job_id"]
         get_booked_call.append(get_booked_for_job_id(job_id))
 
-    booking_rows_for_running_jobs = asyncio.gather(*get_booked_call)
+    booking_rows_for_running_jobs = await asyncio.gather(*get_booked_call, return_exceptions=True)
```
dusktreader: How does the loop below handle gathered exception results?

agent/lm_agent/workload_managers/slurm/cmd_utils.py
```diff
@@ -60,25 +60,39 @@ class LicenseBookingRequest(BaseModel):
     lead_host: str
 
 
-async def get_required_licenses_for_job(slurm_job_id: str) -> List:
+async def get_required_licenses_for_job(slurm_job_id: str, user_name: str, lead_host: str) -> Any:
```
dusktreader: Seems like the return type should be LicenseBookingRequest

agent/lm_agent/workload_managers/slurm/cmd_utils.py
```diff
@@ -60,25 +60,39 @@ class LicenseBookingRequest(BaseModel):
...
 
-    required_features = list()
+    license_booking_request = LicenseBookingRequest(
+        job_id=slurm_job_id,
+        bookings=[],
+        user_name=user_name,
+        lead_host=lead_host,
+    )
+    if not license_array:
+        return []
```
dusktreader: I don't understand why we would return an empty list here. Should it be None instead since a successful lookup returns a single instance of LicenseBookingRequest?

agent/lm_agent/workload_managers/slurm/cmd_utils.py
```diff
@@ -60,25 +60,39 @@ class LicenseBookingRequest(BaseModel):
...
 
     if license_array[0] != "(null)":
         for requested_license in license_array:
             license_regex = re.compile(r"(\w+)\.(\w+)@(\w+):(\d+)")
             if license_regex.match(requested_license):
                 # If the regex matches, parse the values
-                product_feature, _ = requested_license.split("@")
+                product_feature, license_server_tokens = requested_license.split("@")
```
dusktreader: Seems like we should be using the match groups for this instead of using `split()`.

Also, I would like to see this logic split out into a function with a set of unit tests that bash on the edge-cases a little bit

backend/lm_backend/api/license.py
```diff
@@ -83,12 +90,57 @@ async def _get_these_licenses(product_features: Sequence[str]) -> List[LicenseUs
...
 
+async def delete_if_in_use_booking(license: LicenseUseReconcileRequest):
+    """
+    Check the database for the license, if it is booked, then delete it.
+    """
+    queries = []
+    for used_license in license.used_licenses:
+        query = (
+            booking_table.select()
+            .where(booking_table.c.lead_host == used_license["lead_host"])
```
dusktreader, line 101: Nitpicky comment: you can send all three expressions to a single where:

```python
booking_table.select().where(
    booking_table.c.lead_host == used_kicense["lead_host"],
    booking_table.c.user_name == used_kicense["user_name"],
    booking_table.c.booked == used_kicense["booked"],
)
```

backend/lm_backend/api/license.py
```diff
@@ -83,12 +90,57 @@ async def _get_these_licenses(product_features: Sequence[str]) -> List[LicenseUs
...
+            .where(booking_table.c.user_name == used_license["user_name"])
+            .where(booking_table.c.booked == used_license["booked"])
+        )
+        queries.append(database.fetch_one(query))
+    fetched = await asyncio.gather(*queries)
+    bookings = []
+    for item in fetched:
```
dusktreader: maybe collapse into list comprehension

dusktreader (changes requested):

agent/lm_agent/reconciliation.py
zsinx6: To leave it in

jamesbeedy: > do we want to leave this in, or was this only for debugging?

Both :)

agent/lm_agent/reconciliation.py
zsinx6: I'm not sure which way is better for readability, don't know if handling the exception in the outside function is good as well

agent/lm_agent/reconciliation.py
zsinx6: It doesn't, will remove this flag, if any exception happens in this method, something is wrong and should to go sentry

backend/lm_backend/api/license.py
zsinx6: Actually not really:
`TypeError: where() takes 2 positional arguments but 5 were given`, and also I think this is more easy to add things if easy "operation" (e.g. where) is in its own line

agent/lm_agent/reconciliation.py
dusktreader: Fair enough. I think usage will illuminate if we need an exception or not. This is good for now

backend/lm_backend/api/license.py
dusktreader: Welp. The databases docs say they support sqlalchemy API, but it doesn't seem to line up. Good the way it was!

backend/tests/api/test_license.py
```diff
@@ -141,3 +144,114 @@ async def test_map_bookings(some_licenses, insert_objects):
...
+        {"booked": 13, "lead_host": "host1", "user_name": "user1"},
+        {"booked": 14, "lead_host": "host1", "user_name": "user1"},
+    ]
+    license_reconcile_request = LicenseUseReconcileRequest(
+        used=19, product_feature="hello.world", total=100, used_licenses=used_licenses
+    )
+
+    await license._delete_if_in_use_booking(license_reconcile_request)
+
+    booking_rows = [BookingRow.parse_obj(x) for x in await database.fetch_all(booking_table.select())]
```
dusktreader: I don't think we need to parse the response in this test. We could just count the results

backend/tests/api/test_license.py
```diff
@@ -141,3 +144,114 @@ async def test_map_bookings(some_licenses, insert_objects):
...
+    await insert_objects(some_licenses, license_table)
+
+    used_licenses = []
+    license_reconcile_request = LicenseUseReconcileRequest(
+        used=19, product_feature="hello.world", total=100, used_licenses=used_licenses
+    )
+
+    await license._delete_if_in_use_booking(license_reconcile_request)
+
+    booking_rows = [BookingRow.parse_obj(x) for x in await database.fetch_all(booking_table.select())]
```
dusktreader: Same here. No need to parse

backend/tests/api/test_license.py
```diff
@@ -141,3 +144,114 @@ async def test_map_bookings(some_licenses, insert_objects):
...
+    license_reconcile_request = LicenseUseReconcileRequest(
+        used=19, product_feature="hello.world", total=100, used_licenses=used_licenses
+    )
+
+    response = await backend_client.patch(
+        "/api/v1/license/reconcile", json=[license_reconcile_request.dict()]
+    )
+    assert response.status_code == status.HTTP_200_OK
+
+    booking_rows = [BookingRow.parse_obj(x) for x in await database.fetch_all(booking_table.select())]
```
dusktreader: extra parse()

dusktreader (changes requested):
A few more comments for you

agent/lm_agent/workload_managers/slurm/slurmctld_epilog.py
```diff
@@ -42,8 +42,10 @@ async def main():
     # Acqure the job context and get the job_id.
     ctxt = get_job_context()
     job_id = ctxt["job_id"]
+    user_name = ctxt["user_name"]
+    lead_host = ctxt["lead_host"]
 
-    license_booking_request = await get_required_licenses_for_job(job_id)
+    license_booking_request = await get_required_licenses_for_job(job_id, user_name, lead_host)
```
jamesbeedy: Since we don't use `user_name` or `grace_time` in the `get_required_licenses_for_job()` function, possibly it is better to have it return a list of `LicenseBooking` instead of a `LicenseBookingRequest`?

backend/lm_backend/api/license.py
```diff
@@ -83,12 +90,54 @@ async def _get_these_licenses(product_features: Sequence[str]) -> List[LicenseUs
...
+    Check the database for the license, if it is booked, then delete it.
+    """
+    queries = []
+    for used_license in license.used_licenses:
+        query = (
+            booking_table.select()
+            .where(booking_table.c.lead_host == used_license["lead_host"])
+            .where(booking_table.c.user_name == used_license["user_name"])
+            .where(booking_table.c.booked == used_license["booked"])
+            .where(booking_table.c.product_feature == license.product_feature)
```
jamesbeedy, line 104: 👍

jamesbeedy (changes requested):
Looking good! Resolve the comments and then I'll be +1.

agent/lm_agent/reconciliation.py
```diff
@@ -16,18 +16,14 @@
...
-class FailedToRemoveBookedViaGraceTime(Exception):
-    """Unable to delete booking row."""
-
-
 async def remove_booked_for_job_id(job_id: str):
     """
     Send DELETE to /api/v1/booking/book/{job_id}.
     """
-    response = await async_client().delete("/api/v1/booking/book{job_id}")
+    response = await async_client().delete(f"/api/v1/booking/book/{job_id}")
```
matheushent, line 23: shouldn't this be a redirection instead of a direct call?

agent/lm_agent/reconciliation.py
```diff
@@ -44,47 +40,54 @@ async def get_booked_for_job_id(job_id: str) -> Dict:
     """
     Return the booking row for the given job_id.
     """
-    response = await async_client().get("/api/v1/booking/job/{job_id}")
-    book_row = response.json()
-    return book_row
+    response = await async_client().get(f"/api/v1/booking/job/{job_id}")
```
matheushent, line 43: same comment about redirection

agent/lm_agent/reconciliation.py
zsinx6: Can you explain more?

agent/lm_agent/workload_managers/slurm/cmd_utils.py
```diff
@@ -92,9 +108,11 @@ async def check_feature_token_availablity(lbr: LicenseBookingRequest) -> bool:
     with httpx.Client() as client:
         resp = client.get(f"{settings.AGENT_BASE_URL}/api/v1/license/all", headers=LM2_AGENT_HEADERS)
```
matheushent, line 119: same comment about redirection

agent/lm_agent/reconciliation.py
matheushent: I'd like to see `RedirectResponse` from starlette being used here. It seems more coherent to me. Thoughts?

matheushent: It seems to me that doing the current way the API iteself increases its traffic. It seems inefficient some way

agent/lm_agent/workload_managers/slurm/slurmctld_epilog.py
zsinx6: Actually we build the entire license_booking_request using the `user_name` and `lead_host`. If we build the object outside the function it will only move the logic around, I don't see the point of modify this

agent/lm_agent/reconciliation.py
matheushent: > Can you explain more?

The current way the API calls itself, which I think may be inefficient since the API would be increasing its own traffic. Perhaps we could improve it using `RedirectResponse` from starlette

zsinx6: But why redirect? We need to actually make a call to that endpoint

matheushent: yep, but internally it doesn't necessary mean a call but a redirection. I wonder the efficiency of this. We can improve that

matheushent: redirection instead of direct call means less network traffic

zsinx6: No, the api will not call itself, it is going to call the backend

zsinx6: Take a look at the async_client: https://github.com/omnivector-solutions/license-manager/blob/75829ed873bc0f3c1c66902da5a8ba08e7ccc326/agent/lm_agent/forward.py#L25

matheushent: I still think this can be improved to a internal redirection instead of a local network call. But anyway, I'd like more thoughts on that

matheushent: I couldn't see in starlette's code whether this is done the same way

zsinx6: The backend is in AWS, it must be a outside call

matheushent: Perhaps I meant confusedly. I meant not returning the redirect response but using the RedirectResponse just to call the API internally

matheushent: oh now I see, it doesn't make sense what I'm saying. Just ignore it

dusktreader (approved):
Nice work!

agent/lm_agent/workload_managers/slurm/slurmctld_epilog.py
jamesbeedy: I guess I wasn't very clear ... my thoughts are basically that if we don't need the `LicenseBookingRequest` object, we just need the bookings, then why pass this information around and create this `LicenseBookingRequest` when we can just change the return type of the function and have it only return a list of bookings?

jamesbeedy (approved):
Thank you for resolving this! 👍

PR #59, opened 2021-09-15
Booking account

zsinx6:
#### What
Add booking accounting.
Make sure to handle the response correctly and with logs.
Small fixes in .gitignore and Makefile.
Removed dead code.

#### Why
We need to account for the current number of booking before creating a new one.

### DIFF
--- a/agent/Makefile
+++ b/agent/Makefile
@@ -35,3 +35,5 @@ clean-build:
 	@rm -fr build/
 	@rm -fr dist/
 	@rm -fr *.egg-info
+
+SHELL := /bin/bash
--- a/agent/lm_agent/workload_managers/slurm/slurmctld_prolog.py
+++ b/agent/lm_agent/workload_managers/slurm/slurmctld_prolog.py
@@ -17,7 +17,6 @@ from lm_agent.backend_utils import get_config_from_backend
 from lm_agent.logs import init_logging, logger
 from lm_agent.workload_managers.slurm.cmd_utils import (
     LicenseBookingRequest,
-    check_feature_token_availablity,
     get_required_licenses_for_job,
     make_booking_request,
     reconcile,
@@ -81,28 +80,13 @@ async def prolog():
         except Exception as e:
             logger.error(f"Failed to call reconcile with {e}")
             sys.exit(1)
-        # Check that there are sufficient feature tokens for the job.
-        try:
-            feature_token_availability = await check_feature_token_availablity(
-                tracked_license_booking_request
-            )
-        except Exception as e:
-            logger.error(f"Failed to call check_feature_token_availablity with {e}")
-            sys.exit(1)
-        if feature_token_availability:
-            # If we have sufficient tokens for features that are
-            # requested, proceed with booking the tokens for each feature.
-            booking_request = await make_booking_request(tracked_license_booking_request)
-            if booking_request:
-                logger.debug(f"License booking sucessful, job id: {job_id}.")
-                logger.debug(f"Licenses booked: {repr(tracked_licenses)}")
-            else:
-                logger.debug("Booking request unsuccessful.")
-                sys.exit(1)
+        booking_request = await make_booking_request(tracked_license_booking_request)
+        if booking_request:
+            logger.debug(f"License booking sucessful, job id: {job_id}.")
+            logger.debug(f"Licenses booked: {repr(tracked_licenses)}")
         else:
-            logger.debug("Not enough feature tokens for job to proceed.")
+            logger.debug("Booking request unsuccessful.")
             sys.exit(1)
-
     sys.exit(0)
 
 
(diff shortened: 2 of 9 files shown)


agent/Makefile
```diff
@@ -35,3 +35,5 @@ clean-build:
 	@rm -fr build/
 	@rm -fr dist/
 	@rm -fr *.egg-info
+
+SHELL := /bin/bash
```
zsinx6, line 39: For some reason I had problems without this, and it is already there for the backend part. But I'm not sure why it interferes with poetry

heitorPB: Poetry is not POSIX compliant, so you have to force to use Bash instead.

agent/lm_agent/workload_managers/slurm/slurmctld_prolog.py
```diff
@@ -81,28 +80,13 @@ async def prolog():
...
-            else:
-                logger.debug("Booking request unsuccessful.")
-                sys.exit(1)
+        booking_request = await make_booking_request(tracked_license_booking_request)
+        if booking_request:
+            logger.debug(f"License booking sucessful, job id: {job_id}.")
+            logger.debug(f"Licenses booked: {repr(tracked_licenses)}")
         else:
-            logger.debug("Not enough feature tokens for job to proceed.")
+            logger.debug("Booking request unsuccessful.")
```
dusktreader: Should this be `logger.error(...)`?

agent/lm_agent/workload_managers/slurm/slurmctld_prolog.py
```diff
@@ -81,28 +80,13 @@ async def prolog():
...
-            # requested, proceed with booking the tokens for each feature.
-            booking_request = await make_booking_request(tracked_license_booking_request)
-            if booking_request:
-                logger.debug(f"License booking sucessful, job id: {job_id}.")
-                logger.debug(f"Licenses booked: {repr(tracked_licenses)}")
-            else:
-                logger.debug("Booking request unsuccessful.")
-                sys.exit(1)
+        booking_request = await make_booking_request(tracked_license_booking_request)
+        if booking_request:
```
dusktreader: I think you could restructure this without an else as:
```python
if not booking_request:
    logger.error(...)
    sys.exit(1)
logger.debug(success_message)
```

I'm not sure that the explicit sys.exit(0) is necessary. Doesn't python use a 0 by default when a process completes normally?

backend/lm_backend/api/booking.py
```diff
@@ -47,6 +46,37 @@ async def get_config_id_for_product_features(product_feature: str) -> Union[int,
     return config_row.id
 
 
+async def _is_booking_available(booking: Booking):
+    """For each product_feature in the booking.features, check if the total of in_use_total + new_booked
```
dusktreader: The language here is a little bit confusing. Maybe "check if the total available is greater than the sum of in_use_total and new_booked."

backend/lm_backend/api/booking.py
```diff
@@ -58,6 +88,11 @@ async def create_booking(booking: Booking):
 
     An error occurs if the new total for `booked` exceeds `total`
     """
+    if not await _is_booking_available(booking):
+        raise HTTPException(
+            status_code=400,
+            detail=(f"Couldn't book {booking.job_id}, not enough licenses available."),
```
dusktreader: I don't think the parens are necessary.

dusktreader (changes requested):

agent/lm_agent/workload_managers/slurm/slurmctld_prolog.py
zsinx6: I think it is good to be as explicit as possible there

backend/lm_backend/api/booking.py
zsinx6: Can you check again now?

backend/lm_backend/api/booking.py
```diff
@@ -47,6 +46,37 @@ async def get_config_id_for_product_features(product_feature: str) -> Union[int,
...
+    """
+    query = booking_table.select()
+    fetched = await database.fetch_all(query)
+    all_bookings = [BookingRow.parse_obj(x) for x in fetched]
+
+    query = license_table.select()
+    fetched = await database.fetch_all(query)
+    product_feature_licenses = [LicenseUse.parse_obj(x) for x in fetched]
+
+    for product_feature in booking.features:
```
jamesbeedy: The use of product_feature and feature don’t match what the variable actually represents, this makes reading and understanding the code less clear. Can we clean up the terminology so that, for example,  a product_feature is a product_feature and a feature is a feature?

backend/lm_backend/api/booking.py
```diff
@@ -47,6 +46,37 @@ async def get_config_id_for_product_features(product_feature: str) -> Union[int,
...
+        in_use_total = 0
+        total = 0
+        for license in product_feature_licenses:
```
jamesbeedy: This too

dusktreader (approved):
Looks good!

jamesbeedy:
👍 from me. Possibly we wait until after the demo tomorrow to merge this, incase there are any additional adds we may need to make.

PR #64, opened 2021-09-30
Clean bookings

zsinx6:
#### What
Remove not used remaining bookings.

#### Why
To avoid unused bookings.

### DIFF
--- a/agent/lm_agent/backend_utils.py
+++ b/agent/lm_agent/backend_utils.py
@@ -67,7 +67,7 @@ class BackendBookingRow(BaseModel):
     cluster_name: str
 
 
-async def get_bookings_from_backend(cluster_name: str = None) -> List[BackendBookingRow]:
+async def get_bookings_from_backend(cluster_name: Optional[str] = None) -> List[BackendBookingRow]:
     client = async_client()
     bookings: List = []
     try:
--- a/agent/lm_agent/backend_utils.py
+++ b/agent/lm_agent/backend_utils.py
@@ -48,6 +48,44 @@ class BackendConfigurationRow(BaseModel):
     grace_time: int
 
 
+class BackendBookingRow(BaseModel):
+    """
+    NOTE: This is a copy of the schema from the backend.
+          If the schema changes upstream in a non-reverse-compatible
+          way, this schema should cause errors in deserialization.
+    """
+
+    class Config:
+        extra = "ignore"
+
+    job_id: str
+    product_feature: str
+    booked: int
+    config_id: int
+    lead_host: str
+    user_name: str
+    cluster_name: str
+
+
+async def get_bookings_from_backend(cluster_name: str = None) -> List[BackendBookingRow]:
+    client = async_client()
+    bookings: List = []
+    try:
+        if cluster_name:
+            resp = await client.get(f"/api/v1/booking/all?cluster_name={cluster_name}")
+        else:
+            resp = await client.get("/api/v1/booking/all")
+    except ConnectError as e:
+        logger.error(f"Connection failed to backend: {e}")
+        return bookings
+    for booking in resp.json():
+        try:
+            bookings.append(BackendBookingRow.parse_obj(booking))
+        except ValidationError as err:
+            logger.error(f"Wrong format for booking: {str(err)}")
+    return bookings
+
+
 async def get_config_from_backend() -> List[BackendConfigurationRow]:
     client = async_client()
     path = GET_CONFIG_URL_PATH
--- a/agent/lm_agent/backend_utils.py
+++ b/agent/lm_agent/backend_utils.py
@@ -1,7 +1,7 @@
 """
 Provide utilities that communicate with the backend.
 """
-from typing import List
+from typing import List, Optional
 
 from fastapi import status
 from httpx import ConnectError
--- a/agent/lm_agent/reconciliation.py
+++ b/agent/lm_agent/reconciliation.py
@@ -159,6 +159,8 @@ async def reconcile(cluster_name: str = None):
         if slurm_used is None:
             slurm_used = 0
         new_quantity = license_total - license_used - bookings_sum + slurm_used
+        if new_quantity > license_total:
+            new_quantity = license_total
         product, feature = product_feature.split(".")
         update_resource = await sacctmgr_modify_resource(product, feature, new_quantity)
         if update_resource:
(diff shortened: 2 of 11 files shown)


agent/lm_agent/backend_utils.py
```diff
@@ -48,6 +48,44 @@ class Config:
...
+    job_id: str
+    product_feature: str
+    booked: int
+    config_id: int
+    lead_host: str
+    user_name: str
+    cluster_name: str
+
+
+async def get_bookings_from_backend(cluster_name: str = None) -> List[BackendBookingRow]:
```
dusktreader: Seems like this would cause a Mypy error. Probably need to add Optional:

```suggestion
async def get_bookings_from_backend(cluster_name: Optional[str] = None) -> List[BackendBookingRow]:
```

agent/lm_agent/reconciliation.py
```diff
@@ -89,11 +96,52 @@ async def clean_booked_grace_time():
...
+async def clean_bookings(squeue_result, cluster_name):
+    cluster_bookings = [booking.job_id for booking in await get_bookings_from_backend(cluster_name)]
+    jobs_not_running = [job["job_id"] for job in squeue_result if job["state"] != "RUNNING"]
+    all_jobs_squeue = [job["job_id"] for job in squeue_result]
+    delete_booking_call = []
+    for job_id in cluster_bookings:
+        if job_id in jobs_not_running:
+            delete_booking_call.append(remove_booked_for_job_id(job_id))
+        if job_id not in all_jobs_squeue:
+            delete_booking_call.append(remove_booked_for_job_id(job_id))
```
dusktreader: Seems like we could combine the two if clauses:

```suggestion
        if job_id in jobs_not_running or job_id not in all_jobs_squeue:
            delete_booking_call.append(remove_booked_for_job_id(job_id))
```

agent/lm_agent/reconciliation.py
```diff
@@ -89,11 +96,52 @@ async def clean_booked_grace_time():
...
+    for license_data in licenses_to_update:
+        product_feature = license_data["product_feature"]
+        bookings_sum = license_data["bookings_sum"]
+        license_total = license_data["license_total"]
+        license_used = license_data["license_used"]
+        slurm_used = await get_tokens_for_license(product_feature + "@flexlm", "Used")
+        if slurm_used is None:
+            slurm_used = 0
+        new_quantity = license_total - license_used - bookings_sum + slurm_used
+        if new_quantity > license_total:
```
dusktreader, line 142: Maybe use `max()` and `min()` here instead to do a range clamp?
```
new_quantity = min(new_quantity, license_total)
new_quantity = max(new_quantity, 0)
```

You could combine to one line as well for compactness, but it hurts readability.

Just for fun, I saw this on Stack Overflow (should not use because it's confusing, but I thought it was amusing):
```python
new_quantity = sorted((0, new_quantity, license_total))[1]
```

agent/lm_agent/reconciliation.py
```diff
@@ -89,11 +96,52 @@ async def clean_booked_grace_time():
...
+            new_quantity = license_total
+        if new_quantity < 0:
+            new_quantity = 0
+        product, feature = product_feature.split(".")
+        update_resource = await sacctmgr_modify_resource(product, feature, new_quantity)
+        if update_resource:
+            logger.info("Slurmdbd updated successfully.")
+        else:
+            logger.info("Slurmdbd update unsuccessful.")
```
dusktreader, line 151: should this be `logger.error` or `logger.warn`?

agent/lm_agent/workload_managers/slurm/cmd_utils.py
```diff
@@ -252,6 +252,30 @@ async def scontrol_show_lic():
...
+async def get_cluster_name() -> str:
+    cmd = [
+        SACCTMGR_PATH,
+        "list",
+        "cluster",
+        "-nP",
+        "format=Cluster",
+    ]
+    logger.debug("##### sacctmgr get cluster name cmd #####")
+    logger.debug(f"{' '.join(cmd)}")
```
dusktreader, line 264: Seems like you could do the log after the `shlex.join()` so you don't have to do this separately.

agent/lm_agent/workload_managers/slurm/cmd_utils.py
```diff
@@ -252,6 +252,30 @@ async def scontrol_show_lic():
     return output
 
 
+async def get_cluster_name() -> str:
+    cmd = [
+        SACCTMGR_PATH,
+        "list",
+        "cluster",
+        "-nP",
```
dusktreader, line 260: for readability, it might be good to  use the full long-options here.

agent/lm_agent/workload_managers/slurm/slurmctld_epilog.py
```diff
@@ -5,68 +5,58 @@
...
-                else:
-                    logger.info("Slurmdbd update unsuccessful.")
-
-    sys.exit(0)
+        # Attempt to remove the booking and log the result.
+        booking_removed = await _remove_booking_for_job(job_id)
+        if booking_removed:
+            logger.debug(f"Booking for job id: {job_id} successfully deleted.")
+        else:
+            logger.debug(f"Booking for job id: {job_id} not removed.")
```
dusktreader, line 59: Should this be warn or error?

dusktreader (changes requested):
Nice work! Left a few comments for you.

agent/lm_agent/workload_managers/slurm/slurmctld_epilog.py
zsinx6: No, this is just for debug

agent/lm_agent/workload_managers/slurm/cmd_utils.py
zsinx6: It is just a debug log, so I think it doesn't matter much

agent/lm_agent/reconciliation.py
zsinx6: I think we just want to inform this in the logs, so info seems correct

agent/lm_agent/reconciliation.py
zsinx6: We will change this in the next PR

dusktreader (approved):

PR #66, opened 2021-10-18
Fix debug log level

zsinx6:
#### What
Change log level to debug.

#### Why
To avoid bad understood.


dusktreader (approved):

heitorPB (approved):

heitorPB (changes requested):
Please update the CHANGELOG as well

zsinx6:
We need to "create" the changelog properly first, I think it should be another PR, we also need to make the proper "release"

heitorPB (approved):

jamesbeedy (approved):
👍

PR #67, opened 2021-10-22
Remove fastapi from agent

zsinx6:
#### What
The agent is now just a script.

#### Why
To avoid an unnecessary fastapi running server.

https://app.clickup.com/t/1r4rmac


agent/pyproject.toml
```diff
@@ -54,6 +50,7 @@ include_trailing_comma = true
 [tool.poetry.scripts]
 lm-configure = "bin.lmconfigure:app"
 lm-create-jwt = "bin.createjwt:main"
+reconcile = "lm_agent.reconcile:main"
```
heitorPB, line 53: :+1:

agent/lm_agent/backend_utils.py
```diff
@@ -25,7 +24,7 @@ async def get_license_manager_backend_version() -> str:
     """Return the license-manager-backend version."""
     resp = await async_client().get("/version")
     # Check that we have a valid response.
-    if resp.status_code != status.HTTP_200_OK:
+    if resp.status_code != 200:
```
dusktreader, line 27: We should still use named constants. The httpx package has them. Like `httpx.codes.OK`. There are several other places throughout the code this should apply to as well.

agent/lm_agent/reconcile.py
```diff
@@ -0,0 +1,51 @@
...
+    """Main function to setup the env and call the reconcile function."""
+    logger.info("Starting reconcile script")
+    await backend_version_check()
+    begin_logging()
+    await reconcile()
+    logger.info("Reconcile completed successfully")
+
+
+def main():
+    asyncio.run(run_reconcile())
```
dusktreader, line 51: What is the benefit of running this asynchronously?

agent/lm_agent/reconcile.py
```diff
@@ -0,0 +1,51 @@
...
+    level = getattr(logging, settings.LOG_LEVEL)
+    logger.setLevel(level)
+    logger.info(f"Backend URL: {settings.BACKEND_BASE_URL}")
+
+
+async def run_reconcile():
+    """Main function to setup the env and call the reconcile function."""
+    logger.info("Starting reconcile script")
+    await backend_version_check()
+    begin_logging()
```
dusktreader: Shouldn't this go before the `backend_version_check`? There are log lines in that method.

agent/lm_agent/workload_managers/slurm/cmd_utils.py
```diff
@@ -5,15 +5,14 @@
 import subprocess
 from typing import List, Optional, Union
 
-import httpx
 from pydantic import BaseModel, Field
 
-from lm_agent.config import PRODUCT_FEATURE_RX, settings
+from lm_agent.config import PRODUCT_FEATURE_RX
+from lm_agent.forward import async_client
```
dusktreader, line 11: Should we just move `async_client` to this module since we removed the other forwarding logic?

agent/lm_agent/workload_managers/slurm/slurmctld_epilog.py
```diff
@@ -5,29 +5,21 @@
...
-    with httpx.Client() as client:
-        resp = client.delete(
-            f"{settings.AGENT_BASE_URL}/api/v1/booking/book/{job_id}",
-            headers=LM2_AGENT_HEADERS,
-        )
-
-        # Return True if the request to delete the booking was successful.
-        if resp.status_code == 200:
-            return True
+    resp = await async_client().delete(f"api/v1/booking/book/{job_id}")
```
dusktreader, line 19: It seems like this logic should also be moved to the `cmd_utils` module.

agent/lm_agent/reconciliation.py
```diff
@@ -171,25 +171,16 @@ async def update_report():
             "No license data could be collected, check that tools are installed "
             "correctly and the right hosts/ports are configured in settings"
         )
-        raise HTTPException(
-            status_code=status.HTTP_400_BAD_REQUEST,
-            detail="report failed",
-        )
+        raise LicenseManagerBackendConnectionError()
```
dusktreader, line 174: I think it would be useful for a message to be added to this exception that describes the action that failed.

agent/tests/test_reconciliation.py
```diff
@@ -168,7 +168,7 @@ async def test_reconcile_report_empty(report_mock: mock.AsyncMock):
     Check the correct behavior when the report is empty in reconcile.
     """
     report_mock.return_value = []
-    with pytest.raises(HTTPException):
+    with pytest.raises(LicenseManagerBackendConnectionError):
```
dusktreader, line 171: If a message is added (as suggested above), it would be good to use the `match` keyword here to make sure the exception is coming from the expected place.

dusktreader (changes requested):
Nice Work (and quickly done)! Left a few comments for you

agent/lm_agent/backend_utils.py
heitorPB: > We should still use named constants.

Why? I prefer using numbers, as it is explicit to what it means and is also shorter.

zsinx6: Yeah, like there is the https://docs.python.org/3/library/http.html#http.HTTPStatus natively, but I choose to use the numbers, if it had the same format as fastapi (with the number + the name) I would switch

agent/lm_agent/reconcile.py
zsinx6: Nothing, but we need to make async calls inside the function, and it would take a lot more time to change everything to be sync

agent/lm_agent/reconcile.py
zsinx6: Totally, will fix

agent/lm_agent/reconciliation.py
zsinx6: The log above already gives all information necessary

agent/lm_agent/workload_managers/slurm/cmd_utils.py
zsinx6: Maybe move it to backend_utils then, but for now I think we can move with the way it is, and just make sure it is really working together with the charm

jamesbeedy:
+1 once this is tested. Nice work!

dusktreader (approved):

heitorPB (approved):

PR #68, opened 2021-10-28
Add Dockerfile and docker-compose.yaml

julianaklulo:
#### What
Add Dockerfile and docker-compose to build the application and the database.

#### Why
The project didn't have a way to run the application using containers.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -8,9 +8,6 @@ RUN poetry export -f requirements.txt --output requirements.txt --without-hashes
 
 FROM python:3.8
 
-RUN apt-get -y update
-RUN apt-get install -y --fix-missing build-essential cmake libpq-dev
-
 WORKDIR /app
 
 COPY --from=requirements-stage /tmp/requirements.txt /app/requirements.txt
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,0 +1,34 @@
+FROM python:3.8 as requirements-stage
+
+RUN apt-get -y update
+RUN apt-get install -y --fix-missing build-essential cmake libpq-dev
+
+WORKDIR /tmp
+
+
+RUN pip install poetry
+
+
+COPY ./pyproject.toml ./poetry.lock* /tmp/
+
+
+RUN poetry export -f requirements.txt --output requirements.txt --without-hashes
+
+
+FROM python:3.8
+
+
+WORKDIR /app
+
+
+COPY --from=requirements-stage /tmp/requirements.txt /app/requirements.txt
+
+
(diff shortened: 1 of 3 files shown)


zsinx6 (approved):

heitorPB (changes requested):
Could you add some notes to the README (or somewhere else) about these?

backend/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM python:3.8 as requirements-stage
+
+RUN apt-get -y update
+RUN apt-get install -y --fix-missing build-essential cmake libpq-dev
```
heitorPB: Why these dependencies are necessary?

zsinx6: postgres

heitorPB: But we are not building the postgres image here.

The lm app starts with a clean `python:3.8` image, right (see line 18 below)? Shouldn't these deps be there instead?

zsinx6: Totally

julianaklulo: I moved the postgres dependencies to the LM image.

julianaklulo:
> Could you add some notes to the README (or somewhere else) about these?

@heitorPB the README seems to be outdated, but I'll include the command to run the backend using `docker-compose`.

dusktreader:
> > Could you add some notes to the README (or somewhere else) about these?
> 
> @heitorPB the README seems to be outdated, but I'll include the command to run the backend using `docker-compose`.

There's a task to update the README as a part of the V3 deliverables: https://app.clickup.com/t/18022949/LM-43.

Adding some info about the docker-compose is fine for now, and we'll circle back to updating the README after V2.1 is finished.

backend/Dockerfile
julianaklulo: Actually, we don't need these dependencies, I removed them and it worked.

backend/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+FROM python:3.8
+
+WORKDIR /app
+
+COPY --from=requirements-stage /tmp/requirements.txt /app/requirements.txt
+RUN pip install --no-cache-dir --upgrade -r /app/requirements.txt
+
+
+COPY . /app
+RUN pip install .
```
heitorPB, line 18: Do we need two `pip install` commands?

julianaklulo: The `pip install .` is needed for this: https://github.com/omnivector-solutions/license-manager/blob/9a3ec291d74233550c39a28f41838a74e9c88dc3/backend/lm_backend/main.py#L62

PR #69, opened 2021-11-03
Add RLM parser

julianaklulo:
#### What
Add RLM parser and unit test.

#### Why
We need to add suport for RLM license server.

`Task`: https://app.clickup.com/t/18022949/LM-128

### DIFF
--- a/agent/lm_agent/parsing/rlm.py
+++ b/agent/lm_agent/parsing/rlm.py
@@ -54,6 +54,8 @@ def parse(s: str) -> dict:
             )
         elif parsed_data:
             parsed_count = RX_COUNT.match(lines[i + 1])
+            if not parsed_count:
+                continue
             count_data = parsed_count.groupdict()
             data_data = parsed_data.groupdict()
             parsed_dict["total"].append(
--- a/agent/lm_agent/parsing/rlm.py
+++ b/agent/lm_agent/parsing/rlm.py
@@ -1,0 +1,55 @@
+"""
+Parser for rlm
+"""
+import re
+
+INT = r"\d+"
+VERSION = rf"v{INT}\.{INT}"
+HOSTWORD = r"[a-zA-Z0-9-]+"
+HOSTWORD2 = r"[a-zA-Z0-9-.]+"
+HOSTNAME = rf"{HOSTWORD}(\.{HOSTWORD2})*"
+
+TOTALS_LINE = rf"^(?P<license_feature>\w+) {VERSION}: (?P<user_name>\w+)@(?P<lead_host>{HOSTNAME}) (?P<license_used_by_host>\d+).*$"
+DATA_LINE = rf"^\s*(?P<license_feature>\w+) {VERSION}$"
+COUNT_LINE = r"^\s*count: (?P<count>\d+).*inuse: (?P<in_use>\d+).*$"
+
+RX_TOTAL = re.compile(rf"{TOTALS_LINE}")
+RX_DATA = re.compile(rf"{DATA_LINE}")
+RX_COUNT = re.compile(rf"{COUNT_LINE}")
+
+
+def parse(s: str) -> dict:
+    """
+    Parse lines of the license output with regular expressions
+    """
+    parsed_dict: dict = {"uses": [], "total": []}
+    lines = [line.strip() for line in s.splitlines()]
+
+    for i, line in enumerate(lines[45:]):
+        parsed_total = RX_TOTAL.match(line)
+        parsed_data = RX_DATA.match(line)
+        if parsed_total:
+            total_data = parsed_total.groupdict()
+            parsed_dict["uses"].append(
+                {
+                    "user_name": total_data["user_name"],
+                    "lead_host": total_data["lead_host"],
+                    "booked": int(total_data["license_used_by_host"]),
+                    "feature": total_data["license_feature"],
+                }
+            )
+        elif parsed_data:
+            parsed_count = RX_COUNT.match(lines[i + 1])
+            count_data = parsed_count.groupdict()
+            data_data = parsed_data.groupdict()
+            parsed_dict["total"].append(
+                {
+                    "feature": data_data["license_feature"],
+                    "total": int(count_data["count"]),
+                    "used": int(count_data["in_use"]),
+                }
+            )
+        else:
+            continue
+
+    return parsed_dict
--- a/agent/lm_agent/parsing/rlm.py
+++ b/agent/lm_agent/parsing/rlm.py
@@ -13,9 +13,9 @@ TOTALS_LINE = rf"^(?P<license_feature>\w+) {VERSION}: (?P<user_name>\w+)@(?P<lea
 DATA_LINE = rf"^\s*(?P<license_feature>\w+) {VERSION}$"
 COUNT_LINE = r"^\s*count: (?P<count>\d+).*inuse: (?P<in_use>\d+).*$"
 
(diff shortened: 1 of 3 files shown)


agent/lm_agent/parsing/rlm.py
```diff
@@ -0,0 +1,71 @@
...
+TOTALS_LINE = rf"^(?P<license_feature>\w+) {VERSION}: (?P<user_name>\w+)@(?P<lead_host>{HOSTNAME}) (?P<license_used_by_host>\d+).*$"  # noqa
+DATA_LINE = rf"^\s*(?P<license_feature>\w+) {VERSION}$"
+COUNT_LINE = r"^\s*count: (?P<count>\d+).*inuse: (?P<in_use>\d+).*$"
+
+RX_TOTAL = re.compile(rf"{TOTALS_LINE}")
+RX_DATA = re.compile(rf"{DATA_LINE}")
+RX_COUNT = re.compile(rf"{COUNT_LINE}")
+
+
+def _get_start_offset(lines) -> int:
```
dusktreader, line 21: I think it would be good to have a few unit tests for this function specifically. The arithmetic is very specific, and we should hammer on a few edge-cases.

agent/lm_agent/parsing/rlm.py
```diff
@@ -0,0 +1,71 @@
...
+VERSION = rf"v{INT}\.{INT}"
+HOSTWORD = r"[a-zA-Z0-9-]+"
+HOSTWORD2 = r"[a-zA-Z0-9-.]+"
+HOSTNAME = rf"{HOSTWORD}(\.{HOSTWORD2})*"
+
+TOTALS_LINE = rf"^(?P<license_feature>\w+) {VERSION}: (?P<user_name>\w+)@(?P<lead_host>{HOSTNAME}) (?P<license_used_by_host>\d+).*$"  # noqa
+DATA_LINE = rf"^\s*(?P<license_feature>\w+) {VERSION}$"
+COUNT_LINE = r"^\s*count: (?P<count>\d+).*inuse: (?P<in_use>\d+).*$"
+
+RX_TOTAL = re.compile(rf"{TOTALS_LINE}")
```
dusktreader: These are already raw strings. Why do we need to interpolate them with f-strings again?

agent/lm_agent/parsing/rlm.py
```diff
@@ -0,0 +1,71 @@
...
+RX_DATA = re.compile(rf"{DATA_LINE}")
+RX_COUNT = re.compile(rf"{COUNT_LINE}")
+
+
+def _get_start_offset(lines) -> int:
+    """Get the start offset of the license data."""
```
dusktreader, line 22: It would be good to show an example of how the string(s) look to relate to the search logic below.

agent/lm_agent/parsing/rlm.py
```diff
@@ -0,0 +1,71 @@
...
+    return i
+
+
+def parse(s: str) -> dict:
+    """
+    Parse lines of the license output with regular expressions
+    """
+    parsed_dict: dict = {"uses": [], "total": []}
+    start_offset = _get_start_offset(s.splitlines())
+    lines = [line.strip() for line in s.splitlines()[start_offset:]]
```
dusktreader: Should probably only `splitlines()` once.

agent/lm_agent/parsing/rlm.py
```diff
@@ -0,0 +1,71 @@
...
+            data_data = parsed_data.groupdict()
+            parsed_dict["total"].append(
+                {
+                    "feature": data_data["license_feature"],
+                    "total": int(count_data["count"]),
+                    "used": int(count_data["in_use"]),
+                }
+            )
+        else:
+            continue
```
dusktreader: No need for the `else/continue` here.

agent/tests/conftest.py
```diff
@@ -75,3 +75,78 @@ def lm_output():
...
+		obsolete: 0, min_remove: 120, total checkouts: 16
+
+
+	------------------------
+
+	csci license usage status on licser.server.com (port 63133)
+
+	converge_super v3.0: jbemfv@myserver.example.com 29/0 at 11/01 09:01  (handle: 15a) 
+	converge_super v3.0: cdxfdn@myserver.example.com 27/0 at 11/03 10:38  (handle: 128) 
+    converge_super v3.0: jbemfv@myserver.example.com 37/0 at 11/01 09:01  (handle: 15a)
```
dusktreader: indent is off.

agent/tests/conftest.py
```diff
@@ -75,3 +75,78 @@ def lm_output():
         "           jbemfv myserver.example.com /dev/tty (v62.2) (myserver.example.com/24200 12507), "
         "start Thu 10/29 8:09, 37 licenses\n"
     )
+
+
+@fixture
+def rlm_output():
+    return """Setting license file path to 10@licserv.server.com
```
dusktreader, line 82: Instead of having to dedent line 83 and 84, you could do this:
```
return textwrap.dedent("""
    Setting license....
    rlmutil v12.2
    Copyright...


        rlm status...
        rml software....
    ...
    """).strip()
```

This looks better to my eyes for python multiline strings that need to be dedented.

dusktreader (changes requested):

agent/tests/parsing/test_rlm.py
```diff
@@ -0,0 +1,52 @@
+"""
+Test the rlm parser
+"""
+from pytest import mark
+
+from lm_agent.parsing.rlm import parse
+
+
+@mark.parametrize(
```
jamesbeedy, line 9: 👍

agent/tests/conftest.py
heitorPB: The way it is looks simple to read. I see no reason to use `textwrap.dedent` (or similar) here, as it is quite an obscure feature of the language.

agent/lm_agent/parsing/rlm.py
julianaklulo: We added the unit test for this function to check a parseable response and a bad one.

agent/lm_agent/parsing/rlm.py
julianaklulo: We don't need, fixed it.

agent/lm_agent/parsing/rlm.py
julianaklulo: Fixed it.

agent/lm_agent/parsing/rlm.py
julianaklulo: Removed it.

agent/tests/conftest.py
julianaklulo: Fixed it.

agent/lm_agent/parsing/rlm.py
julianaklulo: We have the string in a fixture: https://github.com/omnivector-solutions/license-manager/blob/rlm_parser/agent/tests/conftest.py#L81-L150
It's too big to put as a comment in the function. The idea is that we only need the last two sections of the output, so we use the hyphens to search for the line we want to use as the offset.

agent/tests/conftest.py
julianaklulo: I agree with @heitorPB.

dusktreader: It's just a personal preference. I have the same issue with HEREdocs in other languages. I just don't like lines that aren't indented as part of a function.

dusktreader (approved):
Nice work!

PR #70, opened 2021-11-04
Use RLM parser

julianaklulo:
#### What
Add support for using the RLM parser.

#### Why
We need to support RLM license server.

`Task`: https://app.clickup.com/t/18022949/LM-128

### DIFF
--- a/agent/lm_agent/tokenstat.py
+++ b/agent/lm_agent/tokenstat.py
@@ -12,7 +12,7 @@ from pydantic import BaseModel, Field
 from lm_agent.backend_utils import get_config_from_backend
 from lm_agent.config import ENCODING, PRODUCT_FEATURE_RX, TOOL_TIMEOUT, settings
 from lm_agent.logs import logger
-from lm_agent.parsing import flexlm
+from lm_agent.parsing import flexlm, rlm
 
 
 class LicenseService(BaseModel):
@@ -52,6 +52,27 @@ class LicenseReportItem(BaseModel):
         )
 
 
+class RLMReportItem(BaseModel):
+    """
+    An item in a RLM report, a count of tokens for one product/feature.
+    """
+
+    product_feature: str
+    used: int
+    total: int
+    used_licenses: typing.List
+
+    @classmethod
+    def from_stdout(cls, product, parse_fn, tool_name, stdout):
+        """Create a RLM by parsing the stdout from the program that produced it."""
+        parsed = parse_fn(stdout)
+        return cls(
+            product_feature=f"{product}.{parsed['total']['feature']}",
+            used=parsed["total"]["used"],
+            total=parsed["total"]["total"],
+            used_licenses=parsed["uses"],
+        )
+
 class ToolOptions(BaseModel):
     """
     Specifications for running one of the tools that accesses license servers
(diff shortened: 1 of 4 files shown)


agent/lm_agent/tokenstat.py
```diff
@@ -118,21 +156,27 @@ async def attempt_tool_checks(
         stdout, _ = await asyncio.wait_for(proc.communicate(), TOOL_TIMEOUT)
         output = str(stdout, encoding=ENCODING)
 
-        if proc.returncode == 0:
+        if proc.returncode != 0:
+            logger.warning(f"rc = {proc.returncode}!")
+            logger.warning(output)
```
heitorPB: Could you make these logs a bit more informative?

A return code different than zero here does not mean error?

zsinx6: We just log all information we have (returncode + full cmd output), in the next line we already raise an error

heitorPB: If I see a log message `rc = 7` I'd  be "wtf is this? But fine, it is not an error, just a warning"

dusktreader: I think we should actually include the output and the return code in a single log message. That way, they show up together in log aggregators like Graylog.

agent/lm_agent/tokenstat.py
```diff
@@ -118,21 +156,27 @@ async def attempt_tool_checks(
         stdout, _ = await asyncio.wait_for(proc.communicate(), TOOL_TIMEOUT)
         output = str(stdout, encoding=ENCODING)
 
-        if proc.returncode == 0:
+        if proc.returncode != 0:
+            logger.warning(f"rc = {proc.returncode}!")
```
dusktreader: If we are raising an exception, why are the log levels not ERROR?

agent/tests/test_tokenstat.py
```diff
@@ -45,6 +66,19 @@ def tool_opts() -> tokenstat.ToolOptions:
...
 
 
+@fixture
+def tool_opts_rlm() -> tokenstat.ToolOptions:
+    """
+    A ToolOptions set up for easy testing
+    """
+    return tokenstat.ToolOptions(
+        name="rlm",
+        path=Path(f"{MOCK_BIN_PATH}/rlmstat"),
```
dusktreader: `MOCK_BIN_PATH` is already a `Path`. You can just use the `/` operator:
```
path=MOCK_BIN_PATH / "rlmstat"
```

dusktreader (changes requested):

agent/lm_agent/tokenstat.py
julianaklulo: I changed it to error.

agent/lm_agent/tokenstat.py
julianaklulo: I changed it to show both the output and the return code in the same log message (with level error).

agent/lm_agent/tokenstat.py
```diff
@@ -52,6 +52,38 @@ def from_stdout(cls, product, parse_fn, tool_name, stdout):
         )
 
 
+def _filter_current_feature(parsed_list, product):
+    """
+    Get the current feature from the parsed list.
+    """
+    for feature in parsed_list:
+        if feature["feature"] == product:
+            return feature
```
heitorPB, line 61: This returns the first occurrence of the current feature. We will never have 2 (or more) products in the parsed_list?

jamesbeedy: I'm a little bit confused about this function too ... why are we evaluating this `feature["feature"] == product`?

Possibly it would help to add some clarification in the docstring about what the input arguments should look like or what they should contain and also what the function does with them.

dusktreader (approved):
Nice work!

agent/lm_agent/tokenstat.py
zsinx6: Yes, the list will always have only 1 feature where the feature is equals the `product` passed

jamesbeedy (approved):
👍

PR #72, opened 2021-11-09
Handle empty RLM response use case

julianaklulo:
#### What
Add unit test to ensure that the parser is able to handle the empty response for when there are none licenses checked out.

#### Why
We need to be sure that the parser is handling all use cases.

`Task`: https://app.clickup.com/t/18022949/LM-147


jamesbeedy (approved):
👍

jamesbeedy (approved):
👍

PR #73, opened 2021-11-11
Change RLM parsing to use product.feature instead of product_feature

julianaklulo:
#### What
Change the designation of product and feature to use `.` instead of `_`as separator.

#### Why
To keep the same nomenclature across different license servers, such as flexlm.

`Task`: https://app.clickup.com/t/18022949/LM-148

### DIFF
--- a/agent/tests/test_tokenstat.py
+++ b/agent/tests/test_tokenstat.py
@@ -28,8 +28,8 @@ def one_configuration_row():
 def one_configuration_row_rlm():
     return [
         BackendConfigurationRow(
-            product="converge_super",
-            features={"converge_super": 10},
+            product="converge",
+            features={"super": 10},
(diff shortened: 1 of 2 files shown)


agent/tests/test_tokenstat.py
```diff
@@ -248,27 +248,24 @@ async def test_report_rlm(
     reconcile_list = await tokenstat.report()
     assert reconcile_list == [
```
zsinx6: What do you think about creating more tests for this? With an empty config table and empty used_licenses

julianaklulo: I refactored the test to use parametrize, this way we can test both the case with checked out licenses and the empty one

julianaklulo: Also added another test for empty backend config.

zsinx6 (approved):

PR #74, opened 2021-11-17
Check which configs apply for the cluster

julianaklulo:
#### What
Filter the backend configuration entries before reconciliation by checking which licenses are on the cluster.

#### Why
We were using all the configuration entries in the backend to check the licenses, so if a cluster didn't have them, it would crash. To prevent this, we are filtering the backend configs to check only those that are in the cluster. For this we're using the `scontrol show lic` command to get the cluster's licenses and parsing the output to use as a filter before reconciliation.

`Task`: https://app.clickup.com/t/18022949/LM-139

### DIFF
--- a/agent/lm_agent/tokenstat.py
+++ b/agent/lm_agent/tokenstat.py
@@ -10,7 +10,7 @@ from shlex import quote
 
 from pydantic import BaseModel, Field
 
-from lm_agent.backend_utils import get_config_from_backend
+from lm_agent.backend_utils import BackendConfigurationRow, get_config_from_backend
 from lm_agent.config import ENCODING, PRODUCT_FEATURE_RX, TOOL_TIMEOUT, settings
 from lm_agent.logs import logger
 from lm_agent.parsing import flexlm, rlm
@@ -211,7 +211,7 @@ async def attempt_tool_checks(
         return lri
 
 
-def get_all_product_features_from_cluster(show_lic_output) -> typing.List:
+def get_all_product_features_from_cluster(show_lic_output: str) -> typing.List[str]:
     """
     Returns a list of all product.feature in the cluster
     """
@@ -231,15 +231,17 @@ def get_all_product_features_from_cluster(show_lic_output) -> typing.List:
     return parsed_features
 
 
-def filter_entries_from_backend(entries, my_features):
+def get_local_license_configurations(
+    license_configurations: typing.List[BackendConfigurationRow], local_licenses: typing.List[str]
+):
     """
-    Returns a list entries from the backend that match the features on the cluster
+    Return the license configurations from the backend that are configured on the cluster.
     """
     filtered_entries = []
 
-    for entry in entries:
+    for entry in license_configurations:
         for feature in entry.features.keys():
-            if f"{entry.product}.{feature}" in my_features:
+            if f"{entry.product}.{feature}" in local_licenses:
                 filtered_entries.append(entry)
                 continue
     return filtered_entries
@@ -263,9 +265,9 @@ async def report() -> typing.List[dict]:
 
     # Iterate over the license servers and features appending to list
     # of tools/cmds to be ran.
-    entries = await get_config_from_backend()
-    my_features = get_all_product_features_from_cluster(await scontrol_show_lic())
-    filtered_entries = filter_entries_from_backend(entries, my_features)
+    license_configurations = await get_config_from_backend()
+    local_licenses = get_all_product_features_from_cluster(await scontrol_show_lic())
+    filtered_entries = get_local_license_configurations(license_configurations, local_licenses)
 
     for entry in filtered_entries:
         for license_server_type in tools:
--- a/agent/lm_agent/tokenstat.py
+++ b/agent/lm_agent/tokenstat.py
@@ -215,7 +215,7 @@ def get_all_product_features_from_cluster(show_lic_output) -> typing.List:
     """
     Returns a list of all product.feature in the cluster
     """
-    PRODUCT_FEATURE = "LicenseName=(?P<product>[a-zA-Z0-9]+)[_\-.](?P<feature>\w+)"
+    PRODUCT_FEATURE = r"LicenseName=(?P<product>[a-zA-Z0-9]+)[_\-.](?P<feature>\w+)"
     RX_PRODUCT_FEATURE = re.compile(PRODUCT_FEATURE)
 
     parsed_features = []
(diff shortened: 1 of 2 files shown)


zsinx6 (approved):
👍

agent/lm_agent/tokenstat.py
```diff
@@ -209,6 +211,40 @@ async def attempt_tool_checks(
         return lri
 
 
+def get_all_product_features_from_cluster(show_lic_output) -> typing.List:
+    """
+    Returns a list of all product.feature in the cluster
+    """
+    PRODUCT_FEATURE = r"LicenseName=(?P<product>[a-zA-Z0-9]+)[_\-.](?P<feature>\w+)"
+    RX_PRODUCT_FEATURE = re.compile(PRODUCT_FEATURE)
```
jamesbeedy, line 219: Nice!

agent/lm_agent/tokenstat.py
```diff
@@ -209,6 +211,40 @@ async def attempt_tool_checks(
...
+    Returns a list entries from the backend that match the features on the cluster
+    """
+    filtered_entries = []
+
+    for entry in entries:
+        for feature in entry.features.keys():
+            if f"{entry.product}.{feature}" in my_features:
+                filtered_entries.append(entry)
+                continue
+    return filtered_entries
```
jamesbeedy: +1

agent/lm_agent/tokenstat.py
```diff
@@ -209,6 +211,40 @@ async def attempt_tool_checks(
...
+        if parsed_line:
+            parsed_data = parsed_line.groupdict()
+            product = parsed_data["product"]
+            feature = parsed_data["feature"]
+            parsed_features.append(f"{product}.{feature}")
+
+    return parsed_features
+
+
+def filter_entries_from_backend(entries, my_features):
```
jamesbeedy: I would like to see more meaningful input argument naming here. For example, possibly we could `s/entries/license_configurations/` and `s/my_features/licenses/`?

agent/lm_agent/tokenstat.py
```diff
@@ -209,6 +211,40 @@ async def attempt_tool_checks(
...
+    """
+    Returns a list entries from the backend that match the features on the cluster
```
jamesbeedy:
```suggestion
    Return the license configurations from the backend for licenses that are configured on the cluster.
```

agent/lm_agent/tokenstat.py
```diff
@@ -228,7 +264,10 @@ async def report() -> typing.List[dict]:
     # Iterate over the license servers and features appending to list
     # of tools/cmds to be ran.
     entries = await get_config_from_backend()
-    for entry in entries:
+    my_features = get_all_product_features_from_cluster(await scontrol_show_lic())
```
jamesbeedy: How do you feel about changing the naming from `my_features` to `local_licenses`?

agent/lm_agent/tokenstat.py
```diff
@@ -228,7 +264,10 @@ async def report() -> typing.List[dict]:
     # Iterate over the license servers and features appending to list
     # of tools/cmds to be ran.
     entries = await get_config_from_backend()
```
jamesbeedy:
```suggestion
    license_configurations = await get_config_from_backend()
```

agent/lm_agent/tokenstat.py
```diff
@@ -228,7 +264,10 @@ async def report() -> typing.List[dict]:
...
-    for entry in entries:
+    my_features = get_all_product_features_from_cluster(await scontrol_show_lic())
+    filtered_entries = filter_entries_from_backend(entries, my_features)
```
jamesbeedy:
```suggestion
    filtered_entries = filter_entries_from_backend(license_configurations, local_licenses)
```

agent/lm_agent/tokenstat.py
```diff
@@ -209,6 +211,40 @@ async def attempt_tool_checks(
         return lri
 
 
+def get_all_product_features_from_cluster(show_lic_output) -> typing.List:
```
jamesbeedy: How do you feel about adding a type hint to the input argument?

agent/lm_agent/tokenstat.py
jamesbeedy: How do you feel about this function name instead?
```suggestion
def get_local_license_configurations(license_configurations, local_licenses):
```

jamesbeedy: Type hints?

agent/tests/test_tokenstat.py
```diff
@@ -37,6 +37,22 @@ def one_configuration_row_rlm():
     ]
 
 
+@fixture
+def scontrol_show_lic_output():
+    return """
+LicenseName=testproduct1.feature1@nashlic
```
jamesbeedy:
```suggestion
LicenseName=testproduct1.feature1@rlm
```

agent/tests/test_tokenstat.py
```diff
@@ -37,6 +37,22 @@ def one_configuration_row_rlm():
...
+    Total=10 Used=0 Free=10 Reserved=0 Remote=yes
+    """
+
+
+@fixture
+def scontrol_show_lic_output_rlm():
+    return """
+LicenseName=converge.super@nashlic
```
jamesbeedy:
```suggestion
LicenseName=converge.super@rlm
```

agent/lm_agent/tokenstat.py
jamesbeedy: I would like to see a test for this function that asserts it returns the correct output for 3 different cases; 1) `show_lic_output` contains 0 licenses, 2)  `show_lic_output`  contains 1 license, and 3) `show_lic_output`  contains 2 licenses.

agent/lm_agent/tokenstat.py
jamesbeedy: I know this function is simple, but possibly we should add a test for it simple to keep up with our coding standards and keep the coverage up. Thoughts?

jamesbeedy (changes requested):
Looking really good..... requested a few changes/considerations

agent/lm_agent/tokenstat.py
julianaklulo: Sure, I added a test covering these test cases.

agent/lm_agent/tokenstat.py
julianaklulo: Yep, I added the test.

jamesbeedy (approved):
This looks great! 👍

PR #75, opened 2021-11-25
Fix regex and database URL default value

julianaklulo:
#### What
Fix the regex for parsing product and features from cluster.
Also fix database default URL and bump lm-agent version.

#### Why
The regex was not parsing correctly the features from the cluster.
The default value was causing the project to use a sqlite database instead of the postgres defined in the docker-compose.

### DIFF
--- a/agent/lm_agent/tokenstat.py
+++ b/agent/lm_agent/tokenstat.py
@@ -215,7 +215,7 @@ def get_all_product_features_from_cluster(show_lic_output: str) -> typing.List[s
     """
     Returns a list of all product.feature in the cluster
     """
-    PRODUCT_FEATURE = r"LicenseName=(?P<product>[a-zA-Z0-9]+)[_\-.](?P<feature>\w+)"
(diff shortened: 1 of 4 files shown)


agent/lm_agent/tokenstat.py
```diff
@@ -215,7 +215,7 @@ def get_all_product_features_from_cluster(show_lic_output: str) -> typing.List[s
     """
     Returns a list of all product.feature in the cluster
     """
-    PRODUCT_FEATURE = r"LicenseName=(?P<product>[a-zA-Z0-9]+)[_\-.](?P<feature>\w+)"
+    PRODUCT_FEATURE = r"LicenseName=(?P<product>[a-zA-Z0-9_]+)[_\-.](?P<feature>\w+)"
```
zsinx6, line 218: What about a test to ensure this behavior is now covered?

julianaklulo: Sure, I added a new case in the parametrize of this function's test.

zsinx6 (approved):

PR #77, opened 2021-12-01
Fix path to /version endpoint

julianaklulo:
#### What
Fix path to the /version endpoint, adding the /lm prefix.

#### Why
The lm-agent was not starting because it couldn't get the backend version from the endpoint.


matheushent (approved):
Good catch that I missed. 👍🏻

jamesbeedy:
👏

PR #78, opened 2021-12-01
LM-119 Added security via Armasec

dusktreader:
#### What
This adds authorization on a route-to-route basis via Armasec.

#### Why
This is key to moving License Manager to k8s. Next, we will remove the jawthorizer.

`Task`: https://app.clickup.com/t/18022949/LM-119
---

#### Peer Review

### DIFF
--- a/agent/pyproject.toml
+++ b/agent/pyproject.toml
@@ -32,6 +32,7 @@ pytest-freezegun = "^0.4.2"
 mypy = "^0.910"
 types-click = "^7.1.5"
 flake8 = "^3.9.2"
+ipython = "^7.30.0"
 
 [tool.pytest.ini_options]
 minversion = "6.0"
--- a/agent/pyproject.toml
+++ b/agent/pyproject.toml
@@ -32,7 +32,6 @@ pytest-freezegun = "^0.4.2"
 mypy = "^0.910"
 types-click = "^7.1.5"
 flake8 = "^3.9.2"
-ipython = "^7.30.0"
 
 [tool.pytest.ini_options]
 minversion = "6.0"
--- a/backend/lm_backend/api/license.py
+++ b/backend/lm_backend/api/license.py
@@ -22,7 +22,7 @@ router = APIRouter()
 @router.get(
     "/all",
     response_model=List[LicenseUse],
-    dependencies=[Depends(guard.lockdown("license-manager:in-use:read"))],
+    dependencies=[Depends(guard.lockdown("license-manager:license:read"))],
 )
 async def licenses_all():
     """
@@ -36,7 +36,7 @@ async def licenses_all():
 @router.get(
     "/cluster_update",
     response_model=List[Dict],
-    dependencies=[Depends(guard.lockdown("license-manager:in-use:read"))],
+    dependencies=[Depends(guard.lockdown("license-manager:license:read"))],
 )
 async def licenses_and_bookings_to_update():
     """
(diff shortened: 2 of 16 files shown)


agent/pyproject.toml
```diff
@@ -32,6 +32,7 @@ pytest-freezegun = "^0.4.2"
 mypy = "^0.910"
 types-click = "^7.1.5"
 flake8 = "^3.9.2"
+ipython = "^7.30.0"
```
heitorPB: why `ipython` is here? Is it really necessary?

backend/lm_backend/api/license.py
```diff
@@ -11,14 +11,19 @@
...
 
 PRODUCT_FEATURE_RX = r"^.+?\..+$"
 router = APIRouter()
 
 
-@router.get("/all", response_model=List[LicenseUse])
+@router.get(
+    "/all",
+    response_model=List[LicenseUse],
+    dependencies=[Depends(guard.lockdown("license-manager:in-use:read"))],
```
jamesbeedy: `s/in-use/license/g`

matheushent (changes requested):
The convention for the permission name is quite different from the Admin API. I think we should stardardize it. My suggestion is:

`<operation>:<application>:<target>`

e.g. `write:license-manager:config`, `read:license-manager:config`

dusktreader:
> The convention for the permission name is quite different from the Admin API. I think we should stardardize it. My suggestion is:
> 
> `<operation>:<application>:<target>`
> 
> e.g. `write:license-manager:config`, `read:license-manager:config`

It agrees with the permissions set up in Jobbergate:
`<application>:<target>:<operation>`

agent/pyproject.toml
dusktreader: It's just added here as a dev dependency because I use it for deugging/testing. I can remove it, but it won't affect production builds.

dusktreader: Removed in latest commit.

backend/lm_backend/api/license.py
dusktreader: Changed in latest commit.

backend/tests/api/test_config.py
```diff
@@ -198,32 +322,69 @@ async def test_update_nonexistant_configuration(backend_client: AsyncClient):
...
+
+    inject_security_header("owner1", "license-manager:config:write")
     resp = await backend_client.delete("/lm/api/v1/config/100")
     assert resp.status_code == 200
     assert resp.json()["message"] == "Deleted 100 from the configuration table."
 
 
+@mark.asyncio
+@database.transaction(force_rollback=True)
+async def test_delete_configuration__success(
```
julianaklulo: The name of this test is the same as the test above, it should be
```suggestion
async def test_delete_configuration__fail_on_bad_permission(
```

julianaklulo (changes requested):
This is looking good!
Found one test with an odd name though.

backend/tests/api/test_config.py
dusktreader: Oooh, good catch! There was a lot of copy-pasta being thrown around! Not surprised some of it stuck to the wall.

dusktreader: Fixed in latest commit.

jamesbeedy (approved):
Sweet!!!!! 👍

PR #79, opened 2021-12-02
LM-144 Added machine-to-machine auth for license-manager-agent

dusktreader:
#### What
Added support for machine-to-machine auth via Auth0 and armasec to the agent

#### Why
With the backend using armasec for token security, it is necessary to update the agent so it can consume the backend with the new security scheme.

`Task`: https://app.clickup.com/t/18022949/LM-144

---

#### Peer Review


agent/lm_agent/backend_utils.py
```diff
@@ -1,32 +1,92 @@
...
+    logger.debug(f"Posting Auth0 request to {auth0_url}")
+    response = httpx.post(auth0_url, data=auth0_body)
+    LicenseManagerAuthTokenError.require_condition(
+        response.status_code == 200, f"Failed to get auth token from Auth0: {response.text}"
+    )
+    with LicenseManagerAuthTokenError.handle_errors("Malformed response payload from Auth0"):
+        token = response.json()["access_token"]
+
+    logger.debug("Successfully acquired auth token from Auth0")
+    return token
```
matheushent: as far as I could understand, this token is capable of calling the management API and change stuff there. This is dangerous. Is there another way to do it?

dusktreader: This app is only give access to the regular Auth0 api that standard user tokens have access to as well. That's controlled in Auth0

dusktreader: See: 
image

matheushent: ohhhh I see. Thanks for explaining

agent/bin/lmconfigure.py
```diff
@@ -98,19 +85,15 @@ def update(
     if product:
         ctxt['product'] = product
     if features:
-        ctxt['features'] = features.split(",")
```
julianaklulo, line 101: Why is this not needed anymore?

agent/tests/test_backend_utils.py
```diff
@@ -7,73 +7,72 @@
...
+                dict(
+                    product="SomeProduct",
+                    features={"A": 1, "list": 2, "of": 3, "features": 4},
+                    license_servers=["A", "list", "of", "license", "servers"],
+                    license_server_type="O-Negative",
+                    grace_time=13,
+                ),
+                # Invalid config row
+                dict(bad="Data. Should NOT work"),
+                # Another valid conig row
```
julianaklulo, line 211:
```suggestion
                # Another valid config row
```

backend/docker-compose.yaml
```diff
@@ -10,14 +11,22 @@ services:
       timeout: 5s
       retries: 5
     user: postgres
+
   license-manager:
     build:
       context: .
-      dockerfile: ./Dockerfile
+      dockerfile: ./Dockerfile-dev
```
julianaklulo, line 18: Why do we need a new Dockerfile? And what is the volume being used for?

agent/bin/lmconfigure.py
dusktreader: I don't think it was working in the first place. The backend's config endpoint and table both expect the "features" field to be a string. Splitting it was resulting in a 422 from the backend.

backend/docker-compose.yaml
dusktreader: The dev dockerfile allows us to volume mount the source code so that we can have the uvicorn server restart when one of the files is changed. This way, you don't have to bring down your docker images and spin them back up every time you make a code change.

agent/tests/test_backend_utils.py
dusktreader: Fixed in latest commit.

agent/pyproject.toml
```diff
@@ -48,6 +49,14 @@ src_paths = ["lm_agent", "tests"]
...
+warn_unused_configs = true
+plugins = "pydantic.mypy"
+
+[[tool.mypy.overrides]]
+module = [ "buzz", "pkg_resources" ]
+ignore_missing_imports = true
+
 [tool.poetry.scripts]
 lm-configure = "bin.lmconfigure:app"
 lm-create-jwt = "bin.createjwt:main"
```
jamesbeedy: please remove this and the associated code - thx

backend/lm_backend/api/config.py
```diff
@@ -77,7 +77,10 @@ async def add_configuration(configuration: ConfigurationRow):
     """
     Add a configuration to the database for the first time.
     """
-    query = config_table.insert().values(**vars(configuration))
+    query = config_table.insert().values(
+        # It is necessary to exclude None so the database won't attempt to insert a null id
```
jamesbeedy, line 81: 👍

agent/lm_agent/backend_utils.py
```diff
@@ -1,32 +1,92 @@
...
+    logger.debug("Attempting to acquire token from Auth0 with config")
+    auth0_body = dict(
+        audience=settings.AUTH0_AUDIENCE,
+        client_id=settings.AUTH0_CLIENT_ID,
+        client_secret=settings.AUTH0_CLIENT_SECRET,
+        grant_type="client_credentials",
+    )
+    auth0_url = f"https://{settings.AUTH0_DOMAIN}/oauth/token"
+    logger.debug(f"Posting Auth0 request to {auth0_url}")
+    response = httpx.post(auth0_url, data=auth0_body)
```
jamesbeedy: can we cache the token here and only refresh or regenerate if it is expired?

jamesbeedy (changes requested):
Nice work!

agent/pyproject.toml
dusktreader: Removed in latest commit.

agent/lm_agent/backend_utils.py
dusktreader: Caching added in latest commit (and a bunch of unit tests).

backend/docker-compose.yaml
julianaklulo: Nice, thanks for the explanation!

julianaklulo (approved):
Nice work with the token cache!

PR #82, opened 2021-12-03
Get license's server type from backend config row

julianaklulo:
#### What
Get the license's server type from the backend config row.

#### Why
The server type was hardcoded in the code, causing the `slurm_used` value to be None when the license was from a different server type.


jamesbeedy (approved):
Great!!!

jamesbeedy:
👍

PR #87, opened 2021-12-08
Fix rlmstat command

julianaklulo:
#### What
Change the args for invoking `rlmstat`. 

#### Why
The binary for RLM is actually `rlmutil`, `rlmstat` is one of the tools available.

**Obs:** http://www.reprisesoftware.com/RLM_License_Administration.pdf (page 43 describes the usage of `rlmstat`)


jamesbeedy (approved):
👍

PR #91, opened 2021-12-15
remove the 'LM2_' prefix from the Settings class

matheushent:
#### What
Simply as it is, this PR removes the env prefix from the Settings class for the backend. Note nothing changed in the agent. Why? Because the backend is supposed to run in a container while the agent is not supposed to always run in isolation mode.

#### Why
Needed to remove unnecessary complexity =.

`Task`: https://app.clickup.com/t/18022949/LM-165

---

#### Peer Review


julianaklulo (approved):
Nice!

PR #94, opened 2021-12-20
Raise clear exception when report is empty

julianaklulo:
#### What
Change the exception raised when the report is empty in reconciliation.

#### Why
The exception that was being used before is misleading, because it implies that the empty report is related to backend connection issues, when in fact it happens when there aren't licenses in the backend.

### DIFF
--- a/agent/lm_agent/reconciliation.py
+++ b/agent/lm_agent/reconciliation.py
@@ -9,6 +9,7 @@ from httpx import ConnectError
 
 from lm_agent.backend_utils import (
     LicenseManagerBackendConnectionError,
+    LicenseManagerEmptyReportError,
     get_bookings_from_backend,
     get_config_from_backend,
     get_config_id_from_backend,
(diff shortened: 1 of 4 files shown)


jamesbeedy (approved):
100% - Nice

👍

agent/lm_agent/reconciliation.py
```diff
@@ -173,7 +174,7 @@ async def update_report():
             "No license data could be collected, check that tools are installed "
             "correctly and the right hosts/ports are configured in settings"
         )
-        raise LicenseManagerBackendConnectionError()
+        raise LicenseManagerEmptyReportError()
```
heitorPB, line 177: How/where is this handled?

jamesbeedy: I **think** the goal was to raise an exception with a more meaningful name because the previous exception name was misleading/non-representative of the actual problem happening.

julianaklulo: Exactly. We stumbled in this a few weeks ago and spent a lot of time checking for backend connectivity issues, because of the exception that was being raised, when in fact the problem was that there wasn't any licenses in the cluster that matched the configuration in the backend to be reconciled.

heitorPB (approved):

PR #96, opened 2021-12-27
Tucker/lm 164  update release process

dusktreader:
#### What
Added release scripts for nextgen License Manager.

#### Why
Re-using and re-tooling the release process for Jobbergate will make releasing new versions easier and standardized.

This will also prepare us to update the documentation once it's been merged and ported from the main branch.

`Task`: https://app.clickup.com/t/18022949/LM-164

---

#### Peer Review

### DIFF
--- a/release-scripts/check-prereqs.sh
+++ b/release-scripts/check-prereqs.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+
+#BASE_BRANCH="armada-mvp"
+BASE_BRANCH="tucker/LM-164--update-release-process"
+SUBPROJECT=${PWD##*/}
+ALL_SUBPROJECTS="agent backend cli"
+
+function bail {
+    echo "!!! $1 -- Aborting !!!"
+    echo
+    help
+    echo "  * You must have git configured with a signing key"
+    echo "  * You must have the $BASE_BRANCH branch checked out"
+    echo "  * You must have no staged changes"
+    echo "  * Your $BASE_BRANCH branch must be up-to-date with origin"
+    echo "  * You must have no local or remote tags matching the target version"
+    echo "  * You must run the script from one of the sub-projects: $ALL_SUBPROJECTS"
+    exit 1
+}
+
+echo "Checking prerequisites"
+
+echo "Checking if current directory is a releaseable subproject"
+if [[ " ${ALL_SUBPROJECTS[@]} " =~ " ${SUBPROJECT} " ]]
+then
+    echo "$SUBPROJECT is a valid sub-project"
+else
+    bail "This script must be run from one of the valid sub-projects: $ALL_SUBPROJECTS"
+fi
+
+echo "Checking for $BASE_BRANCH branch"
+current_branch=$(git rev-parse --abbrev-ref HEAD)
+if [[ $current_branch -ne "$BASE_BRANCH" ]]
+then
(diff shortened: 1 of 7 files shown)


release-scripts/check-prereqs.sh
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
+
+BASE_BRANCH="armada-mvp"
```
julianaklulo, line 3: Why is the branch name hardcoded?

release-scripts/publish-release.sh
```diff
@@ -0,0 +1,25 @@
...
+function help {
+    echo "Usage: $0"
+}
+
+source $SCRIPTS_HOME/check-prereqs.sh
+
+version_target=$(poetry version --short)
+
+echo "Creating git tag for release"
+git tag --sign --message="Release $version_target" $version_target
```
julianaklulo: The tag is for the whole project, right? So how can we identify if the release is in the agent or in the backend?

julianaklulo (commented):
This is very nice, will save us a lot of time!
Just have a few comments to address.

release-scripts/check-prereqs.sh
dusktreader: This branch will be changed to "main" after legacy LM is decommissioned. All releases should be based off of the main branch. I could make this read from an environment variable, tho.

release-scripts/publish-release.sh
dusktreader: This is something I've struggled with thinking about. With Jobbergate and License Manager both being sort of mono-repos, we can't really create a release with tags that is isolated to one of the sub-projects. That's why I opted for prefixing the tag. I'd love to come up with an alternative if we can make it work with the current structure of these projects.

julianaklulo: The message would be "Release x.y.z", since `poetry version` displays only the x.y.z part. What do you think about making it "Release agent x.y.z" or "Release backend x.y.z"? Even though the tag would apply to the whole repo, this way we could identify to which project the tag is related to.

release-scripts/check-prereqs.sh
julianaklulo: After we get rid of the armada-mvp branch, I guess it'll make sense to hardcode 'main' in the script. Not sure if it's worth adding the env var.

release-scripts/publish-release.sh
dusktreader: Great suggestion! Will update

dusktreader: addressed in latest commit

release-scripts/publish-release.sh
```diff
@@ -0,0 +1,25 @@
...
+source $SCRIPTS_HOME/check-prereqs.sh
+
+version_target=$(poetry version --short)
+
+echo "Creating git tag for release"
+git tag --sign --message="Release $SUBPROJECT $version_target" $SUBPROJECT-$version_target
+
+
+echo "Pushing tag to origin"
+git push origin $version_target
```
julianaklulo: Since the tag is now `$SUBPROJECT-$version_target`, shoudn't this be like this?
```suggestion
git push origin $SUBPROJECT-$version_target
```

dusktreader: fixed in latest commit

julianaklulo (approved):

PR #100, opened 2022-01-11
Fix lmstat command

julianaklulo:
#### What
Fix lmstat path to use the correct lmutil binary.

#### Why
The binary we were using is named incorrectly.


jamesbeedy (approved):
👍

PR #105, opened 2022-01-17
Fix regex expression for token matching

julianaklulo:
#### What
Fix regex expression to assume the quantity of licenses needed for a job is 1 when no token quantity is provided.

#### Why
Slurm assumes the quantity is 1 when it's not provided, so we should also accept it.

`Task`: https://app.clickup.com/t/18022949/LM-175

### DIFF
--- a/agent/lm_agent/workload_managers/slurm/cmd_utils.py
+++ b/agent/lm_agent/workload_managers/slurm/cmd_utils.py
@@ -59,12 +59,17 @@ class LicenseBookingRequest(BaseModel):
 
 
 def _match_requested_license(requested_license: str) -> Union[dict, None]:
-    license_regex = re.compile(r"(?P<product>\w+)\.(?P<feature>\w+)@(?P<server_type>\w+):(?P<tokens>\d+)")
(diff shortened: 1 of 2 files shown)


agent/lm_agent/workload_managers/slurm/cmd_utils.py
```diff
@@ -59,12 +59,17 @@ class LicenseBookingRequest(BaseModel):
 
 
 def _match_requested_license(requested_license: str) -> Union[dict, None]:
-    license_regex = re.compile(r"(?P<product>\w+)\.(?P<feature>\w+)@(?P<server_type>\w+):(?P<tokens>\d+)")
+    license_regex = re.compile(r"(?P<product>\w+)\.(?P<feature>\w+)@(?P<server_type>\w+)?(:(?P<tokens>\d+))?")
```
dusktreader: server_type is optional as well as tokens? It seem slike maybe there's an extra `?` here.

dusktreader (changes requested):

agent/lm_agent/workload_managers/slurm/cmd_utils.py
julianaklulo: There was indeed an extra `?`, thanks!
I addressed it in the latest commit.

dusktreader (approved):

jamesbeedy:
🔥

PR #107, opened 2022-01-25
Refactor Tokenstat

julianaklulo:
#### What
Refactor the Tokenstat module.
Each license server now has its own interface for connecting and parsing the server's ouput.
The logic for getting license's report was improved.
The tests were refactored as well.

#### Why
The old code was not scalable for adding support to more license servers.

`Task`: https://app.clickup.com/t/1hptz9b


jamesbeedy:
@julianaklulo will you make the first line of the docstring a proper sentence by ending with a period? Is there something we can put in the flake8 config to enforce this?

dusktreader:
> @julianaklulo will you make the first line of the docstring a proper sentence by ending with a period? Is there something we can put in the flake8 config to enforce this?

We could use fake8-docstrings: https://pypi.org/project/flake8-docstrings/

agent/lm_agent/config.py
```diff
@@ -67,10 +67,10 @@ class Settings(BaseSettings):
     LICENSE_SERVER_FEATURES_CONFIG_PATH: Optional[str]
 
     # path to the binary for lmstat (needed for FlexLM licenses)
-    LMSTAT_BIN_PATH: DirectoryPath = _DEFAULT_BIN_PATH
+    LMUTIL_PATH: DirectoryPath = _DEFAULT_BIN_PATH
```
dusktreader, line 70: It strikes me that settings like this could easily be captured in the config rows in the database. We should not act on this now, but it may be something to consider for later.

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
 import typing
-from pathlib import Path
-from shlex import quote
 
 from pydantic import BaseModel, Field
 
-from lm_agent.backend_utils import BackendConfigurationRow, get_config_from_backend
+from lm_agent.backend_utils import (
+    BackendConfigurationRow,
+    LicenseManagerNonSupportedServerTypeError,
```
dusktreader: These custom exception types should probably live in their own module like `lm_agent.exceptions`.

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
-    A license service such as "flexlm", with a set of host-port tuples
-    representing the network location where the service is listening.
-    """
+class LicenseManagerBadServerOutput(Exception):
+    """Exception for license server bad output."""
 
-    name: str
-    hostports: typing.List[typing.Tuple[str, int]]
 
+class LicenseServerInterface(metaclass=abc.ABCMeta):
```
dusktreader: There should be a docstring for this class.

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
+        )
 
-class LicenseReportItem(BaseModel):
-    """
-    An item in a LicenseReport, a count of tokens for one product/feature
-    """
+    @abc.abstractclassmethod
+    def get_output_from_server(self, product_feature: str):
+        """Return output from license server for the indicated features."""
+        raise NotImplementedError
```
dusktreader: For a method decorated with `abstractclassmethod`, you do not need to explicitly raise `NotImplementedError`. You can just leave it at the docstring.

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
 
-    tool_name: str
-    product_feature: str = Field(..., regex=PRODUCT_FEATURE_RX)
-    used: int
-    total: int
-    used_licenses: typing.List
+    @abc.abstractclassmethod
+    def get_report_item(self, product_feature: str):
+        """Parse license server output into a report item for the indicated feature."""
+        raise NotImplementedError
```
dusktreader: See comment above.

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
+    """Extract license information from FlexLM license server."""
+
+    def __init__(self, license_servers: typing.List[str]):
+        self.license_servers = license_servers
+        self.parser = flexlm.parse
+
+    def get_commands_list(self):
+        """Generate a list of commands with the available license server hosts."""
+
+        host_ports = [(server.split(":")[1], server.split(":")[2]) for server in self.license_servers]
```
dusktreader: We shouldn't need to call `split` twice here.

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
+        # get the list of commands for each license server host
+        commands_to_run = self.get_commands_list()
+
+        # run each command in the list, one at a time, until one succeds
+        for cmd in commands_to_run:
+            feature = product_feature.split(".")[1]
+            cmd = cmd + f" {feature}"
+            proc = await asyncio.create_subprocess_shell(
+                cmd, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.STDOUT
+            )
```
dusktreader:
```suggestion
            feature_cmd = f"{cmd} {feature}"
            proc = await asyncio.create_subprocess_shell(
                feature_cmd, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.STDOUT
            )
```

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
+            return output
+
+    async def get_report_item(self, product_feature: str):
+        """Override abstract method to parse FlexLM license server output into License Report Item."""
+
+        server_output = await self.get_output_from_server(product_feature)
+        parsed_output = self.parser(server_output)
+
+        # raise exception if parser didn't output license information
+        if not parsed_output["total"]:
```
dusktreader: Should we use `get()` here instead or can we trust that the parser will always return output with a "total" item?

```suggestion
        if not parsed_output.get("total"):
```

dusktreader: Should we also check for "uses" and make sure that the "total" sub-dict includes its own "used" and "total"?

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
+        """Generate a list of commands with the available license server hosts."""
 
-def _cleanup_features(features_list):
-    """
-    Remove the feature key for each entry in the parsed["uses"], since we already handled it.
-    """
-    for feature in features_list:
-        del feature["feature"]
-    return features_list
+        host_ports = [(server.split(":")[1], server.split(":")[2]) for server in self.license_servers]
```
dusktreader: See comment above about `split`.

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
-    """
+        # get the list of commands for each license server host
+        commands_to_run = self.get_commands_list()
 
-    tool_name: str
-    product_feature: str
-    used: int
-    total: int
-    used_licenses: typing.List
+        # run each command in the list, one at a time, until one succeds
```
dusktreader, line 141: Is this comment accurate? It seems like if one of the command fails we will raise an exception immediately.

agent/lm_agent/tokenstat.py
```diff
@@ -1,219 +1,223 @@
...
+        current_feature_item = self._filter_current_feature(
+            parsed_output["total"], product_feature.split(".")[1]
+        )
+        feature_booked_licenses = self._filter_used_features(
+            parsed_output["uses"], product_feature.split(".")[1]
+        )
+        used_licenses = self._cleanup_features(feature_booked_licenses)
+
+        # raise exception if parser didn't output license information
+        if not current_feature_item:
```
dusktreader: Should we also check `feature_booked_licenses` and `used_licenses`?

agent/lm_agent/tokenstat.py
```diff
@@ -259,32 +263,31 @@ async def report() -> typing.List[dict]:
...
+        product_features_to_check = []
+        for feature in entry.features.keys():
+            product_feature = f"{entry.product}.{feature}"
+            product_features_to_check.append(product_feature)
+
+        if entry.license_server_type == "flexlm":
+            license_server_interface = FlexLMLicenseServer(entry.license_servers)
+        elif entry.license_server_type == "rlm":
+            license_server_interface = RLMLicenseServer(entry.license_servers)
         else:
```
dusktreader: As we add more license server types, it might be good to create a mapping (dict), attempt to fetch the appropriate type from it, and raise the exception if it cannot be found:

```suggestion
        server_type_map = dict(
            flexlm=FlexLMLicenseServer,
            rlm=RLMLicenseServer,
        )
        server_type = server_type_map.get(entry.license_server_type)
        if server_type is None:
            raise LicenseManagerNonSupportedServerTypeError()
        license_server_interface = server_type(entry.license_servers)
```

dusktreader (changes requested):
Nice work! Made some comments for you. Let me know when it's ready for another look.

agent/lm_agent/tokenstat.py
julianaklulo: Totally, I created the `exceptions` module and moved all the exceptions.

agent/lm_agent/tokenstat.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/tokenstat.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/tokenstat.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/tokenstat.py
julianaklulo: Yes. The license servers are defined like this: `flexlm:licserv0011.com:24200`, so I changed it to `server.split(":")[1:]` to get only the host and port we need and ignore the server type in the beginning. Thanks!

agent/lm_agent/tokenstat.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/tokenstat.py
julianaklulo: Addressed it.

agent/lm_agent/tokenstat.py
julianaklulo: Nice, I addressed in the latest commit. Now it looks much better!

agent/lm_agent/tokenstat.py
julianaklulo: It should be, since we have to try all the hosts in the config row until one of them returns what we expect.
I added a continue statement to check the next host in case one of them fails. If they all fail, then we'll raise the exception.

agent/lm_agent/config.py
julianaklulo: I agree. The problem I see with adding it to the config row would be if we have two different environments using the same backend. The location of the binary in each of them could be different. But it would be better than using env vars for this.

agent/lm_agent/tokenstat.py
julianaklulo: The `flexlm` parser will always return the `total` and `uses`.
If nothing was matched in the output, the `total` will be `None` and the `uses` an empty list.
If the total is not None, then it will always have the `total` and `used` inside.

But I'll add the validations anyway, better play it safe.

agent/lm_agent/tokenstat.py
julianaklulo: The `feature_booked_licenses` can be an empty list (if no one is using the license in the license server), which makes `used_licenses` an empty list as well.

I'll simplify it to only `current_feature_item` and `used_licenses`, since they're the data we need for the LicenseReportItem, and also validate both.

agent/lm_agent/config.py
dusktreader: Yes. We will have to be thinking about this for Q2.

dusktreader (approved):
Looks very good! Nice work.

PR #108, opened 2022-01-27
Tucker/job 121  simplify permissions

dusktreader:
#### What
Instead of maintaining 4 (or 5) different permission types for all the license manager entities, we condensed it down to to two apiece: view or edit

#### Why
This will drastically reduce the number of permissions that we need to track in our apps and our OIDC provider. It's also most likely that permissions that fine-grained wouldn't even be used anyway.

`Task`: https://app.clickup.com/t/18022949/JOB-121

---

#### Peer Review

### DIFF
--- a/backend/lm_backend/api/booking.py
+++ b/backend/lm_backend/api/booking.py
@@ -7,6 +7,7 @@ from fastapi import APIRouter, Depends, HTTPException, Query
 from sqlalchemy.sql import delete, join, select
 
 from lm_backend.api.config import get_config_id_for_product_features
+from lm_backend.api.permissions import Permissions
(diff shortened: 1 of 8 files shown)


backend/lm_backend/api/booking.py
```diff
@@ -7,6 +7,7 @@
 from sqlalchemy.sql import delete, join, select
 
 from lm_backend.api.config import get_config_id_for_product_features
+from lm_backend.api.permissions import Permissions
```
heitorPB, line 10: Did you forget to commit this file? Where is this `Permissions` class defined?

dusktreader: Man, I totally did. Hurrying to get the PR out before I had to leave for the day, and did something dumb. Thanks! (addressed in latest commit)

julianaklulo (approved):

PR #109, opened 2022-01-27
Fix validation errors in Tokenstat

julianaklulo:
#### What
Fix validation for server output, adding unit test for the case when `0` licenses are in use.
Also fix type of the env var for the binaries path.

#### Why
After testing the `tokenstat` refactor, I caught some validations erros, such as: when the feature in the `flexlm` output is not in use `(total == 0)`, it was not passing the validation.

### DIFF
--- a/agent/lm_agent/config.py
+++ b/agent/lm_agent/config.py
@@ -66,11 +66,11 @@ class Settings(BaseSettings):
     # path to the license server features config file
     LICENSE_SERVER_FEATURES_CONFIG_PATH: Optional[str]
 
-    # path to the binary for lmstat (needed for FlexLM licenses)
-    LMUTIL_PATH: DirectoryPath = _DEFAULT_BIN_PATH
+    # path to the binary for lmutil (needed for FlexLM licenses)
+    LMUTIL_PATH: Path = _DEFAULT_BIN_PATH
 
-    # path to the binary for rlmstat (needed for RLM licenses)
-    RLMUTIL_PATH: DirectoryPath = _DEFAULT_BIN_PATH
+    # path to the binary for tlmutil (needed for RLM licenses)
+    RLMUTIL_PATH: Path = _DEFAULT_BIN_PATH
 
     # debug mode turns on certain dangerous operations
     DEBUG: bool = False
--- a/agent/lm_agent/config.py
+++ b/agent/lm_agent/config.py
@@ -4,7 +4,7 @@ from enum import Enum
 from pathlib import Path
 from typing import Optional
 
-from pydantic import BaseSettings, DirectoryPath, Field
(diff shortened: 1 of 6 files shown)


heitorPB (approved):

agent/lm_agent/config.py
```diff
@@ -66,11 +66,11 @@ class Settings(BaseSettings):
...
     LICENSE_SERVER_FEATURES_CONFIG_PATH: Optional[str]
 
-    # path to the binary for lmstat (needed for FlexLM licenses)
-    LMUTIL_PATH: DirectoryPath = _DEFAULT_BIN_PATH
+    # path to the binary for lmutil (needed for FlexLM licenses)
+    LMUTIL_PATH: Path = _DEFAULT_BIN_PATH
 
-    # path to the binary for rlmstat (needed for RLM licenses)
-    RLMUTIL_PATH: DirectoryPath = _DEFAULT_BIN_PATH
+    # path to the binary for tlmutil (needed for RLM licenses)
```
jamesbeedy:
```suggestion
    # path to the binary for rlmutil (needed for RLM licenses)
```

agent/tests/conftest.py
```diff
@@ -82,6 +82,24 @@ def lm_output():
     )
 
 
+@fixture
+def lm_output_no_licenses():
```
jamesbeedy:
```suggestion
def lmstat_output_no_licenses():
```

agent/tests/test_tokenstat.py
```diff
@@ -194,6 +194,21 @@ async def test_rlm_get_report_item_with_no_used_licenses(
     )
 
 
+@mark.asyncio
+@mock.patch("lm_agent.tokenstat.FlexLMLicenseServer.get_output_from_server")
+async def test_flexlm_get_report_item_with_no_used_licenses(
+    get_output_from_server_mock: mock.MagicMock, flexlm_server, lm_output_no_licenses
+):
+    get_output_from_server_mock.return_value = lm_output_no_licenses
```
jamesbeedy:
```suggestion
    get_output_from_server_mock.return_value = lmstat_output_no_licenses
```

agent/lm_agent/config.py
julianaklulo: Addressed in latest commit.

agent/tests/conftest.py
julianaklulo: I renamed all the `lm` features to `lmstat` in the latest commit.

agent/tests/test_tokenstat.py
julianaklulo: I renamed all the `lm` features to `lmstat` in the latest commit.

PR #116, opened 2022-02-09
Fix cache token permission error

julianaklulo:
#### What
Fix permission error when accesssing cached token dir.

#### Why
It was making the prolog to fail.


dusktreader (approved):

jamesbeedy (approved):
👍

PR #117, opened 2022-02-10
Remove License Server Interfaces from tokenstat module

julianaklulo:
#### What
Remove License Server Interfaces from tokenstat, creating modules for each one.
Also add debugging logs to tokenstat.

#### Why
All the license servers were cluttered in tokenstat, making it difficult to add a new interface.

### DIFF
--- a/agent/lm_agent/server_interfaces/flexlm.py
+++ b/agent/lm_agent/server_interfaces/flexlm.py
@@ -1,3 +1,4 @@
+"""FlexLM license server interface."""
 import typing
 
 from lm_agent.config import settings
--- a/agent/lm_agent/server_interfaces/flexlm.py
+++ b/agent/lm_agent/server_interfaces/flexlm.py
@@ -33,7 +33,7 @@ class FlexLMLicenseServer(LicenseServerInterface):
 
         # run each command in the list, one at a time, until one succeds
         for cmd in commands_to_run:
-            feature = product_feature.split(".")[1]
+            (_, feature) = product_feature.split(".")
             feature_cmd = f"{cmd} {feature}"
             output = await run_command(feature_cmd)
 
--- a/agent/lm_agent/server_interfaces/flexlm.py
+++ b/agent/lm_agent/server_interfaces/flexlm.py
@@ -51,11 +51,12 @@ class FlexLMLicenseServer(LicenseServerInterface):
         parsed_output = self.parser(server_output)
 
         # raise exception if parser didn't output license information
-        if (
-            parsed_output.get("total") is None
-            or parsed_output.get("uses") is None
-            or parsed_output.get("total", {}).get("used") is None
-            or parsed_output.get("total", {}).get("total") is None
+        if parsed_output.get("total") is None or any(
+            [
+                parsed_output.get("total", {}).get("used") is None,
+                parsed_output.get("total", {}).get("total") is None,
+                parsed_output.get("uses") is None,
+            ]
         ):
             raise LicenseManagerBadServerOutput("Invalid data returned from parser.")
 
--- a/agent/lm_agent/server_interfaces/flexlm.py
+++ b/agent/lm_agent/server_interfaces/flexlm.py
@@ -1,0 +1,68 @@
+import typing
+
+from lm_agent.config import settings
+from lm_agent.exceptions import LicenseManagerBadServerOutput
+from lm_agent.parsing import flexlm
+from lm_agent.server_interfaces.license_server_interface import LicenseReportItem, LicenseServerInterface
+from lm_agent.server_interfaces.utils import run_command
+
+
+class FlexLMLicenseServer(LicenseServerInterface):
+    """Extract license information from FlexLM license server."""
+
+    def __init__(self, license_servers: typing.List[str]):
+        self.license_servers = license_servers
+        self.parser = flexlm.parse
+
+    def get_commands_list(self):
+        """Generate a list of commands with the available license server hosts."""
+
+        host_ports = [(server.split(":")[1:]) for server in self.license_servers]
+        commands_to_run = []
+        for host, port in host_ports:
+            command_line = f"{settings.LMUTIL_PATH} lmstat -c {port}@{host} -f"
+            commands_to_run.append(command_line)
+        return commands_to_run
+
+    async def get_output_from_server(self, product_feature: str):
+        """Override abstract method to get output from FlexLM license server."""
+
+        # get the list of commands for each license server host
+        commands_to_run = self.get_commands_list()
+
+        # run each command in the list, one at a time, until one succeds
+        for cmd in commands_to_run:
+            feature = product_feature.split(".")[1]
+            feature_cmd = f"{cmd} {feature}"
+            output = await run_command(feature_cmd)
+
+            # try the next server if the previous didn't return the expected data
+            if output is None:
+                continue
+            return output
+
+        raise RuntimeError("None of the checks for FlexLM succeeded!")
+
+    async def get_report_item(self, product_feature: str):
+        """Override abstract method to parse FlexLM license server output into License Report Item."""
+
+        server_output = await self.get_output_from_server(product_feature)
+        parsed_output = self.parser(server_output)
+
+        # raise exception if parser didn't output license information
+        if (
+            parsed_output.get("total") is None
+            or parsed_output.get("uses") is None
+            or parsed_output.get("total", {}).get("used") is None
+            or parsed_output.get("total", {}).get("total") is None
+        ):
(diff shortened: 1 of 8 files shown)


agent/lm_agent/server_interfaces/flexlm.py
```diff
@@ -0,0 +1,68 @@
+import typing
```
dusktreader, line 2: Add a module docstring, please.

agent/lm_agent/server_interfaces/flexlm.py
```diff
@@ -0,0 +1,68 @@
...
+
+    async def get_output_from_server(self, product_feature: str):
+        """Override abstract method to get output from FlexLM license server."""
+
+        # get the list of commands for each license server host
+        commands_to_run = self.get_commands_list()
+
+        # run each command in the list, one at a time, until one succeds
+        for cmd in commands_to_run:
+            feature = product_feature.split(".")[1]
```
dusktreader: Instead of using direct indexing, you *could* do this as:
```suggestion
            (_, feature) = product_feature.split(".")
```

agent/lm_agent/server_interfaces/flexlm.py
```diff
@@ -0,0 +1,68 @@
...
+        server_output = await self.get_output_from_server(product_feature)
+        parsed_output = self.parser(server_output)
+
+        # raise exception if parser didn't output license information
+        if (
+            parsed_output.get("total") is None
+            or parsed_output.get("uses") is None
+            or parsed_output.get("total", {}).get("used") is None
+            or parsed_output.get("total", {}).get("total") is None
+        ):
```
dusktreader:
```suggestion
        if any([
            parsed_output.get("total") is None,
            parsed_output.get("uses") is None,
            parsed_output.get("total", {}).get("used") is None,
            parsed_output.get("total", {}).get("total") is None,
        ]):
```

agent/lm_agent/server_interfaces/license_server_interface.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    Abstract class for License Server interface.
+
+    The logic for obtaining the data output from the License Server should be encapsulated in
+    the get_output_from_server method.
+
+    After obtaining the output, the parsing and manipulation of the data should be implement in
+    the get_report_item method.
+
+    It is expected the license information to be parsed into an LicenseReportItem.
```
dusktreader:
```suggestion
    Abstract base class for License Server interface.

    The logic for obtaining the data output from the License Server should be encapsulated in
    the ``get_output_from_server`` method.

    After obtaining the output, the parsing and manipulation of the data should be implemented in
    the ``get_report_item`` method.

    The license information should be parsed into a ``LicenseReportItem``.
```

agent/lm_agent/server_interfaces/license_server_interface.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @classmethod
+    def __subclasshook__(cls, subclass):
+        return (
+            hasattr(subclass, "get_output_from_server")
+            and callable(subclass.get_output_from_server)
+            and hasattr(subclass, "get_report_item")
+            and callable(subclass.get_report_item)
+            or NotImplemented
+        )
```
dusktreader:
```suggestion
        return all([
            hasattr(subclass, "get_output_from_server"),
            callable(subclass.get_output_from_server),
            hasattr(subclass, "get_report_item"),
            callable(subclass.get_report_item),
        ]) or NotImplemented
```

agent/lm_agent/server_interfaces/rlm.py
```diff
@@ -0,0 +1,96 @@
+import typing
```
dusktreader, line 2: Please add a module docstring.

agent/lm_agent/server_interfaces/rlm.py
```diff
@@ -0,0 +1,96 @@
...
+    def _filter_current_feature(self, parsed_list, feature):
+        """
+        Get the current feature from the parsed list.
+        """
+        for feature_item in parsed_list:
+            if feature_item["feature"].count("_") == 0:  # `converge`
+                if feature_item["feature"] == feature:
+                    return feature_item
+            elif "".join(feature_item["feature"].split("_")[1:]) == feature:
+                # `converge_super` | `converge_gui_polygonica` | `converge_...`
```
dusktreader: I think it would be good to move these comments into the method docstring and explain the two cases with examples.

agent/lm_agent/server_interfaces/rlm.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _filter_current_feature(self, parsed_list, feature):
+        """
+        Get the current feature from the parsed list.
+        """
+        for feature_item in parsed_list:
+            if feature_item["feature"].count("_") == 0:  # `converge`
+                if feature_item["feature"] == feature:
+                    return feature_item
+            elif "".join(feature_item["feature"].split("_")[1:]) == feature:
```
dusktreader: Should there maybe be an `else` clause here that throws an exception if both patterns fail?

agent/lm_agent/server_interfaces/utils.py
```diff
@@ -0,0 +1,21 @@
+import asyncio
```
dusktreader, line 2: Please add a module docstring.

agent/lm_agent/tokenstat.py
```diff
@@ -275,6 +67,12 @@ async def report() -> typing.List[dict]:
     local_licenses = get_all_product_features_from_cluster(await scontrol_show_lic())
     filtered_entries = get_local_license_configurations(license_configurations, local_licenses)
 
+    logger.debug("#### Getting reconciliation report ####")
```
dusktreader, line 70: Nice!

agent/tests/server_interfaces/test_flexlm.py
```diff
@@ -0,0 +1,97 @@
+from unittest import mock
```
dusktreader, line 3: Please add a module docstring.

dusktreader (changes requested):
Added some comments and suggestions for you. Nice work!

agent/lm_agent/server_interfaces/flexlm.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/server_interfaces/flexlm.py
julianaklulo: Nice, thanks!

agent/lm_agent/server_interfaces/flexlm.py
julianaklulo: Nice, thanks!
The only problem is that the `any` is trying to evaluate all conditions, so it the first one ("total" is None) is True, the last two will raise an attribute error due to None not having the `.get` method.
Apparently, the old condition using `or` was not evaluating the rest of the conditions when the first one was True, thus making the test pass.
I combined both of them to make it work correctly.

agent/lm_agent/server_interfaces/license_server_interface.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/server_interfaces/license_server_interface.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/server_interfaces/rlm.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/server_interfaces/rlm.py
julianaklulo: I don't think so, since it will always have `0`, `1` or more underscores in the `feature` name.

julianaklulo: Also, if the feature is not in the list, the method will return None and an exception will be raised by the validation in the `get_report_item` method.

agent/lm_agent/server_interfaces/rlm.py
julianaklulo: Sure, I added a more detailed explanation in the docstring.

agent/lm_agent/server_interfaces/utils.py
julianaklulo: Addressed in latest commit.

agent/tests/server_interfaces/test_flexlm.py
julianaklulo: Addressed in latest commit.

dusktreader (approved):

PR #118, opened 2022-02-14
Add LS-Dyna support

julianaklulo:
#### What
Add a license server interface for LS-Dyna license server.

#### Why
We need to support this license server.

`Task`: https://app.clickup.com/t/18022949/LM-169

---

#### Peer Review


agent/tests/parsing/test_lsdyna.py
```diff
@@ -0,0 +1,57 @@
+"""
+Test the LS-Dyna parser
```
matheushent, line 2: I think we're missing tests for the exceptions, aren't we?

agent/lm_agent/parsing/lsdyna.py
```diff
@@ -0,0 +1,87 @@
...
+def parse(s: str) -> dict:
+    """
+    Parse lines of the license output with regular expressions
+    """
+    parsed_data: dict = {}
+    group_used: int = 0
+
+    lines = s.splitlines()
+
+    for i, line in enumerate(lines):
```
dusktreader: This parser is complex enough with precise regexes that I think we need some more targeted unit tests and some functional decomposition in this method.

I would like to see unit tests that use fixed strings, pass them through the parser, and verify correct output or error conditions.

agent/lm_agent/parsing/lsdyna.py
```diff
@@ -0,0 +1,87 @@
...
+                        "booked": int(usage_data["used"]),
+                    }
+                )
+
+        parsed_total_usage = RX_TOTAL.match(line)
+        if parsed_total_usage is None:
+            continue
+
+        total_usage_data = parsed_total_usage.groupdict()
+        group_used = int(total_usage_data["used"])
```
dusktreader: Should there be a `break` here? We should only have one line that matches `parsed_total_usage`, correct? It's not clear in the code that this is the termination condition, though.

agent/lm_agent/parsing/lsdyna.py
```diff
@@ -0,0 +1,87 @@
...
+
+    for program in parsed_data.keys():
```
dusktreader, line 147: Are we certain that program is always unique?

agent/lm_agent/server_interfaces/lsdyna.py
```diff
@@ -0,0 +1,64 @@
...
+from lm_agent.exceptions import LicenseManagerBadServerOutput
+from lm_agent.parsing import lsdyna
+from lm_agent.server_interfaces.license_server_interface import LicenseReportItem, LicenseServerInterface
+from lm_agent.server_interfaces.utils import run_command
+
+
+class LSDynaLicenseServer(LicenseServerInterface):
+    """Extract license information from LS-Dyna license server."""
+
+    def __init__(self, license_servers: typing.List[str]):
```
dusktreader, line 14: missing docstring.

jamesbeedy (approved):
👍

agent/tests/parsing/test_lsdyna.py
julianaklulo: We're testing the 3 possible outputs the license server can return to us: correct data, data with no licenses in use, and unparseable data.
I can't think of other test cases to add here.

dusktreader (changes requested):

agent/lm_agent/server_interfaces/lsdyna.py
heitorPB: I think we should add `flake8-docstrings` to this and all other projects. It would help catch these situations.

agent/tests/parsing/test_lsdyna.py
matheushent: if refer myself to situations like:

```python
def example():
    raise Exception("foo")

def test_example():
    with pytest.raises(Exception):
        example()
```

I think we're missing unit tests that checks if the right exceptions are raised in case some conditions match

agent/lm_agent/parsing/lsdyna.py
julianaklulo: It should be, why would they have licenses with the same name in the same license server?

agent/lm_agent/server_interfaces/lsdyna.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/parsing/lsdyna.py
julianaklulo: This line is always the last one in the output, but I'll add a `break` there to indicate explicitly that it's the termination condition.

agent/lm_agent/parsing/lsdyna.py
julianaklulo: I refactored the parser and decomposed the `parser` function, so now each regex is being unit tested.

agent/lm_agent/parsing/lsdyna.py
```diff
@@ -0,0 +1,150 @@
...
+    rf"(?P<max>{INT})\s\|\s+"
+    rf"(?P<queue>{INT})"
+)
+
+RX_PROGRAM = re.compile(PROGRAM_LINE)
+RX_USAGE = re.compile(USAGE_LINE)
+RX_TOTAL = re.compile(GROUP_LINE)
+
+
+def parse_program_line(line: str):
```
dusktreader, line 41: Beautiful!

agent/lm_agent/parsing/lsdyna.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Also, all the licenses in the server will have their ``used`` value
+    filled with the group used value, since it'll reflect the correct
+    amount of free licenses in the pool.
+    """
+    parsed_data: dict = {}
+    program_list: list = []
+    group_used: int = 0
+
+    for line in s.splitlines():
```
dusktreader, line 127: This is great! It's exactly what I was hoping for. Nice work.

agent/tests/parsing/test_lsdyna.py
```diff
@@ -0,0 +1,138 @@
...
+    The line contains:
+    - program
+    - expiration date
+    - used CPUs
+    - free CPUs
+    - max CPUs
+    - queue
+    From these, we only need to extract ``program`` and ``max CPUs``.
+    """
+    assert parse_program_line("MPPDYNA          12/30/2022          0    500    500 |     0") == {
```
dusktreader, line 20: Yesss!

dusktreader (approved):
Excellent work! Thanks for taking the feedback and applying it well.

jamesbeedy:
This looks great @julianaklulo ! Thank you!

jamesbeedy:
🦅

PR #121, opened 2022-02-21
LM-178 Fixed token caching mechanism to use home directory

dusktreader:
#### What
Adjusted token caching code to use the current user's home directory.

#### Why
The agent was having issues with root and the slurm user both trying to use the same token.

`Task`: https://app.clickup.com/t/18022949/LM-178

---

#### Peer Review

### DIFF
--- a/agent/lm_agent/backend_utils.py
+++ b/agent/lm_agent/backend_utils.py
@@ -1,6 +1,7 @@
 """
 Provide utilities that communicate with the backend.
 """
+import pathlib
 import typing
 
 import httpx
@@ -12,6 +13,20 @@ from lm_agent.exceptions import LicenseManagerAuthTokenError, LicenseManagerBack
 from lm_agent.logs import logger
 
 
+def _get_cache_dir() -> pathlib.Path:
+    """
+    Get the path to the cache dir. Will always be a sub-directory of the current user's home.
+    """
+    return pathlib.Path.home() / f".cache/license-manager"
+
+
+def _get_token_path() -> pathlib.Path:
+    """
+    Get the path to the access token.
+    """
+    return _get_cache_dir() / "access_token"
+
+
 def _load_token_from_cache() -> typing.Union[str, None]:
     """
     Looks for and returns a token from a cache file (if it exists).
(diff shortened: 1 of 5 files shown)


agent/lm_agent/backend_utils.py
```diff
@@ -12,6 +13,20 @@
 from lm_agent.logs import logger
 
 
+def _get_cache_dir() -> pathlib.Path:
+    """
+    Get the path to the cache dir. Will always be a sub-directory of the current user's home.
+    """
+    return pathlib.Path.home() / f".cache/license-manager"
```
heitorPB: LM user is a system user, not a normal one. `$HOME` might  not be defined and we might not have permissions to create this folder, specially in a shared environment with network mounted `/home`.

Why changing from `/run`?

dusktreader: Addressed this issue in the latest commit. CACHE_DIR restored as an environment variable. No logic to create missing directories. If the cache dir doesn't exist, you don't get cached tokens.

agent/lm_agent/config.py
```diff
@@ -11,6 +11,7 @@
 
 
 DEFAULT_DOTENV_PATH = Path("/etc/default/license-manager-agent")
+DEFAULT_CACHE_DIR = Path.home() / Path(".cache/license-manager")
```
heitorPB, line 14:
```suggestion
DEFAULT_CACHE_DIR = Path("/var/cache/license-manager")
```
?

dusktreader: No, it needs to default to home for running locally. In the charm, it is set to `/var/cache/license-manager`

heitorPB: It took me some time to understand how the code does that. Could you simplify it to make it explicit? There's no need to have two variables to hold the cache directories. This is a huge source of confusion right now.

dusktreader: I don't quite understand. The `DEFAULT_CACHE_DIR` is just the default setting. The final value will either be the same or can be overriden by setting the environment.

julianaklulo (approved):

PR #122, opened 2022-03-02
Add LM-X support

julianaklulo:
#### What
Add a license server interface for LM-X license server.

#### Why
We need to support this license server.

`Task`: https://app.clickup.com/t/18022949/LM-170
---

#### Peer Review

### DIFF
--- a/agent/lm_agent/parsing/lmx.py
+++ b/agent/lm_agent/parsing/lmx.py
@@ -30,8 +30,7 @@ def parse_feature_line(line: str):
     if parsed_feature is None:
         return None
     feature_data = parsed_feature.groupdict()
-    feature_data["feature"] = feature_data["feature"]
-
+    
     return feature_data["feature"]
 
 
--- a/agent/lm_agent/parsing/lmx.py
+++ b/agent/lm_agent/parsing/lmx.py
@@ -1,0 +1,114 @@
+"""
+Parser for LM-X
+"""
+import re
+
+HOSTWORD = r"[a-zA-Z0-9-]+"
+HOSTNAME = rf"{HOSTWORD}(\.{HOSTWORD})*"
+INT = r"\d+"
+VERSION = rf"{INT}\.{INT}"
+
+FEATURE_LINE = rf"^Feature: (?P<feature>\w+) Version: (?P<version>{VERSION}) Vendor: (?P<vendor>\S+)$"
+IN_USE_LINE = rf"^(?P<in_use>{INT}) of (?P<total>{INT}) license\(s\) used(:)?$"
+USAGE_LINE = rf"^(?P<in_use>{INT}) license\(s\) used by (?P<user>\S+)@(?P<lead_host>{HOSTNAME}) \[{INT}\.{INT}\.{INT}\.{INT}\]"
+
+RX_FEATURE = re.compile(FEATURE_LINE)
+RX_IN_USE = re.compile(IN_USE_LINE)
+RX_USAGE = re.compile(USAGE_LINE)
+
+
+def parse_feature_line(line: str):
+    """
+    Parse the feature line in the LM-X output.
+    Data we need:
+    - ``feature``: license name
+    """
+    parsed_feature = RX_FEATURE.match(line)
+    if parsed_feature is None:
+        return None
+    feature_data = parsed_feature.groupdict()
+    feature_data["feature"] = feature_data["feature"].lower()
+
+    return feature_data["feature"]
+
+
+def parse_in_use_line(line: str):
+    """
+    Paser the in use line in the LM-X output.
+    Data we need:
+    - ``in_use``: quantity of licenses being use
+    - ``total``: total amount of licenses
+
+    Obs: this line doesn't include the license name.
+    The license in use is the last one parsed before this line.
+    """
+
+    parsed_in_use = RX_IN_USE.match(line)
+    if parsed_in_use is None:
+        return None
+    in_use_data = parsed_in_use.groupdict()
+
+    return {
+        "used": int(in_use_data["in_use"]),
+        "total": int(in_use_data["total"]),
+    }
+
+
+def parse_usage_line(line: str):
+    """
+    Parse the usage line in the LS-Dyna output.
+    Data we need:
+    -``user_name``: user who booked the license
+    -``lead_host``: host using the license
+    -``booked``: quantity of licenses being used
+
+    Obs: this line also doesn't incluse the license name.
+    The license in use is the last one parsed before this line.
+    """
+    parsed_usage = RX_USAGE.match(line)
+    if parsed_usage is None:
+        return None
+    usage_data = parsed_usage.groupdict()
+
+    return {
+        "user_name": usage_data["user"],
+        "lead_host": usage_data["lead_host"],
+        "booked": int(usage_data["in_use"]),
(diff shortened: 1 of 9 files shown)


agent/lm_agent/parsing/lmx.py
```diff
@@ -0,0 +1,117 @@
...
+    rf"^(?P<in_use>{INT}) license\(s\) "
+    rf"used by (?P<user>\S+)@(?P<lead_host>{HOSTNAME}) "
+    rf"\[{INT}\.{INT}\.{INT}\.{INT}\]"
+)
+RX_FEATURE = re.compile(FEATURE_LINE)
+RX_IN_USE = re.compile(IN_USE_LINE)
+RX_USAGE = re.compile(USAGE_LINE)
+
+
+def parse_feature_line(line: str):
```
dusktreader: I think typehints for the return types here would be good.

agent/lm_agent/parsing/lmx.py
```diff
@@ -0,0 +1,117 @@
...
+    Data we need:
+    - ``feature``: license name
+    """
+    parsed_feature = RX_FEATURE.match(line)
+    if parsed_feature is None:
+        return None
+    feature_data = parsed_feature.groupdict()
+    feature_data["feature"] = feature_data["feature"]
+
+    return feature_data["feature"]
```
dusktreader: This is confusing. What's going on here?

agent/lm_agent/parsing/lmx.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def parse_in_use_line(line: str):
+    """
+    Paser the in use line in the LM-X output.
```
dusktreader:
```suggestion
    Parse the in use line in the LM-X output.
```

agent/lm_agent/parsing/lmx.py
```diff
@@ -0,0 +1,117 @@
...
+    usage_data = parsed_usage.groupdict()
+
+    return {
+        "user_name": usage_data["user"],
+        "lead_host": usage_data["lead_host"],
+        "booked": int(usage_data["in_use"]),
+    }
+
+
+def parse(s: str):
```
dusktreader: It would be good to have a more descriptive parameter name than `s`.

agent/lm_agent/parsing/lmx.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    Parse the LM-X output, using regex to match the lines we need:
```
dusktreader:
```suggestion
    Parse the LM-X output using regex to match the lines we need:
```

agent/lm_agent/parsing/lmx.py
```diff
@@ -0,0 +1,117 @@
...
+    -``feature line``: info about each license
+    -``in use line``: info about licenses in use
+    -``usage line``: info about users using licenses
+
+    Since the in use and usage line don't have the name of the license in use,
+    we're saving each parsed license in a list. This way, we can find which
+    license uthe user is using by checking the last parsed license in the list.
```
dusktreader:
```suggestion
    license the user is using by checking the last parsed license in the list.
```
Also "user is using" could be worded better.

agent/lm_agent/parsing/lmx.py
```diff
@@ -0,0 +1,117 @@
...
+        elif parsed_in_use:
+            feature = feature_list[-1]
+            parsed_data[feature] = {
+                "used": parsed_in_use["used"],
+                "total": parsed_in_use["total"],
+                "uses": [],
+            }
+        elif parsed_usage:
+            feature = feature_list[-1]
+            parsed_data[feature]["uses"].append(parsed_usage)
```
dusktreader, line 115: There's no `else` case here. Can this conditional ever have each previous expression resolve to `False`?

agent/lm_agent/server_interfaces/lmx.py
```diff
@@ -0,0 +1,65 @@
...
+
+    async def get_output_from_server(self):
+        """Override abstract method to get output from LM-X license server."""
+
+        # get the list of commands for each license server host
+        commands_to_run = self.get_commands_list()
+
+        # run each command in the list, one at a time, until one succeds
+        for cmd in commands_to_run:
+            output = await run_command(cmd)
```
dusktreader, line 37: Would this be a useful place to use `asyncio.gather`?

agent/tests/parsing/test_lmx.py
```diff
@@ -0,0 +1,131 @@
+"""
+Test the LM-X parser
+"""
+
+from lm_agent.parsing.lmx import parse, parse_feature_line, parse_in_use_line, parse_usage_line
+
+
+def test_parse_feature_line():
```
dusktreader, line 8: These are great!

dusktreader (changes requested):
Nice work! Added a few comments/suggestions for you.

agent/lm_agent/parsing/lmx.py
julianaklulo: Oh, I forgot to remove this line.
I was using `.lower()` in the feature name before I realized that the LM-X licenses have non-consistent casing.
Thanks for pointing this!

agent/lm_agent/parsing/lmx.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/parsing/lmx.py
julianaklulo: Totally, I renamed it to server_output in the latest commit.

agent/lm_agent/parsing/lmx.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/parsing/lmx.py
julianaklulo: I changed the wording, is it better now?
```suggestion
    license is being used by checking the last parsed license in the list.
```

agent/lm_agent/parsing/lmx.py
julianaklulo: Yes. The license server output contains a lot of lines with information we don't need, so we just ignore them. I'll add a else here to explicitly skip those lines, is it ok?

agent/lm_agent/server_interfaces/lmx.py
julianaklulo: The idea here is to use whichever license server host responds first. Ideally, it will be the first one in the list of hosts (they are mirrors, for redundancy), but we check the other ones just in the case of the first failing to respond.
Do you think `asyncio.gather` is applicable in this situation? I don't think so because we'd run this more times than needed in the case where the first server responds.

agent/lm_agent/parsing/lmx.py
julianaklulo: Addressed in latest commit.

agent/lm_agent/server_interfaces/lmx.py
dusktreader: Oh, that makes sense. In this case, resolving the coroutines in order synchronously makes a lot of sense. Thanks for clarifying.

dusktreader (approved):

PR #135, opened 2022-04-18
Raise exception if Sentry DSN is not set

julianaklulo:
#### What
Raise an exception when the Sentry DSN is not set in the agent charm config.

#### Why
To make it easier to identify when the service crashes because of Sentry DSN not set.

`Task`: https://app.clickup.com/t/18022949/ARMADA-462

---

#### Peer Review

### DIFF
--- a/agent/lm_agent/reconcile.py
+++ b/agent/lm_agent/reconcile.py
@@ -7,15 +7,19 @@ import sentry_sdk
 
 from lm_agent.backend_utils import check_backend_health
 from lm_agent.config import settings
+from lm_agent.exceptions import LicenseManagerSentryDsnNotSet
 from lm_agent.logs import init_logging, logger
 from lm_agent.reconciliation import reconcile
 
-if settings.SENTRY_DSN:
-    sentry_sdk.init(
-        dsn=settings.SENTRY_DSN,
-        sample_rate=typing.cast(float, settings.SENTRY_SAMPLE_RATE),  # The cast silences mypy
-        environment=settings.DEPLOY_ENV,
-    )
+
+if settings.SENTRY_DSN is None:
+    raise LicenseManagerSentryDsnNotSet("Sentry DSN must be set.")
+
+sentry_sdk.init(
+    dsn=settings.SENTRY_DSN,
+    sample_rate=typing.cast(float, settings.SENTRY_SAMPLE_RATE),  # The cast silences mypy
+    environment=settings.DEPLOY_ENV,
+)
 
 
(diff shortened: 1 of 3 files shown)


agent/lm_agent/reconcile.py
```diff
@@ -7,15 +7,18 @@
...
 from lm_agent.reconciliation import reconcile
 
-if settings.SENTRY_DSN:
-    sentry_sdk.init(
-        dsn=settings.SENTRY_DSN,
-        sample_rate=typing.cast(float, settings.SENTRY_SAMPLE_RATE),  # The cast silences mypy
-        environment=settings.DEPLOY_ENV,
-    )
+if settings.SENTRY_DSN is None:
+    raise LicenseManagerSentryDsnNotSet("Sentry DSN must be set.")
```
jamesbeedy, line 15: do we want to always force this? even when you are running it for dev?

heitorPB:
> To make it easier to identify when the service crashes because of Sentry DSN not set.

This bothers me. If Sentry DSN is not set, the service should work without using Sentry instead of crashing. Is there a reason to make LM not work without Sentry?

julianaklulo:
> > To make it easier to identify when the service crashes because of Sentry DSN not set.
> 
> This bothers me. If Sentry DSN is not set, the service should work without using Sentry instead of crashing. Is there a reason to make LM not work without Sentry?

It was crashing when the config was not set because then it would pass a string "None" to the agent. For local development, I usually set the config as an empty string and it works without the Sentry integration.

dusktreader:
> > > To make it easier to identify when the service crashes because of Sentry DSN not set.
> > 
> > 
> > This bothers me. If Sentry DSN is not set, the service should work without using Sentry instead of crashing. Is there a reason to make LM not work without Sentry?
> 
> It was crashing when the config was not set because then it would pass a string "None" to the agent. For local development, I usually set the config as an empty string and it works without the Sentry integration.

After thinking more about this, I think we should allow the environment variable to be unset. Then, we should only bind to sentry if it is set and is a valid URL (not empty or a literal 'None'). This would allow for local development without having to hack around the initialization.

In this case, we could also make the config for the charm optional and the other PR would work as implemented.

julianaklulo:
I'll close this PR because we agreed that the best case here would make this config defaults to an empty string in the agent charm: https://github.com/omnivector-solutions/charm-license-manager-agent/pull/36

This way, if no Sentry DSN is set in the charm, the Sentry integration won't be initialized in the agent.

PR #136, opened 2022-04-27
Fix license config row filtering

julianaklulo:
#### What
Fix config row filtering in tokenstat.

#### Why
If the config row has more than one feature added in the cluster, the entry is being duplicated.
Breaking the inner for after appending the entry fixes the duplication.

---

#### Peer Review


jamesbeedy (approved):
👍

PR #138, opened 2022-05-03
Add filtering to cluster update response during reconciliation

julianaklulo:
#### What
Create a function to filter the `cluster_update` endpoint response to only update licenses available in the cluster.
Also move function to get local licenses to the `cmd_utils module`, since it interacts with scontrol.

#### Why
The reconciliation was trying to update licenses that don't exist in the cluster.

`Task`: https://app.clickup.com/t/18022949/ARMADA-516

---

#### Peer Review


jamesbeedy (approved):
👍

PR #146, opened 2022-05-25
Add client_id column to config table

julianaklulo:
#### What
Add the new `client_id` column to the config table.

#### Why
In order to connect each configuration row to a specific cluster, we need a new column to store the cluster identification (`client_id`) to store which cluster the configuration applies to.

`Task`: https://app.clickup.com/t/18022949/ARMADA-468

---

#### Peer Review


fschuch (approved):
Nice work!

For me the PR is aproved as it is, I just have one suggestion. I'd rather `cluster_name` than `client_id` because it is more expressive and improves readability. Client looks more generic and is not so clear to people that are not so familiar with this project.

julianaklulo:
> I'd rather `cluster_name` than `client_id` because it is more expressive and improves readability. Client looks more generic and is not so clear to people that are not so familiar with this project.

I agree that `cluster_name` is a more intuitive name (we already use this in the bookings table), but according to @dusktreader, after the Keycloak migration we'll use `client_id` to identify where the requests are coming from, so we'll follow this standard across all projects. In this case, the `client_id` for the configuration row in LM will identify the cluster where the config applies to, which will include more than just the cluster's name.

fschuch:
No problem, consistency across projects is also important.

PR #149, opened 2022-06-02
Add route to retrieve configs filtering by client_id

julianaklulo:
#### What
Create a new route `/agent/all` to list configs of a specific cluster by filtering the configs using the client_id in the access token.

#### Why
We need a way to identify if the request is coming from an agent and filter the configs to return only those that belong to the agent's cluster. If we need to fetch all config rows (for the frontend, for example), the old `/all` still works. The new `/agent/all` route should be used only by the agents.

`Task`: https://app.clickup.com/t/18022949/ARMADA-469

---

#### Peer Review

### DIFF
--- a/backend/CHANGELOG.rst
+++ b/backend/CHANGELOG.rst
@@ -7,7 +7,7 @@ This file keeps track of all notable changes to license-manager-backend
 Unreleased
 ----------
 * Add cluster_id column to config table to identify which cluster the configuration applies to
-* Created new route to fetch configurations filtering by the cluster requesting the data
+* Added new route to fetch all configurations from a specific cluster
 
 2.2.7 -- 2022-05-10
 -------------------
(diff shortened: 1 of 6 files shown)


backend/CHANGELOG.rst
```diff
@@ -7,6 +7,7 @@ This file keeps track of all notable changes to license-manager-backend
 Unreleased
 ----------
 * Add cluster_id column to config table to identify which cluster the configuration applies to
+* Created new route to fetch configurations filtering by the cluster requesting the data
```
heitorPB: I am not sure what this means. Is the following correct?
```suggestion
* Added new route to fetch all configuration from a specific cluster
```

julianaklulo: Yes. This route is going to be used only by the agents. It identifies which cluster is requesting the data and automatically filters the configs.

julianaklulo: I changed it in the latest commit to use your suggestion, thanks!

dusktreader (approved):

PR #153, opened 2022-06-29
Changed DEPLOY_ENV setting to a string

dusktreader:
#### What
Changed it to a string so that it can assume any value


julianaklulo (commented):
Can you add this to the changelog, please?

PR #165, opened 2022-09-06
add comment about upgrading the database to revision 5308f26e8b60

matheushent:
#### What
As the name suggests, this PR only adds a comment about one of the alembic migrations. The problem we faced when migrating the database is, due to the nature of the column `client_id`, there are infinite possibilities for its values (theoretically). So, the solutions was to add a dummy value as default and then apply a manual change.

#### Why
Complete the deployment

---

#### Peer Review


julianaklulo (approved):
Thank you, @matheushent!

PR #174, opened 2022-11-07
Add limit to booking licenses

julianaklulo:
#### What
Changes the configuration row to include a limit in the feature information. When creating a configuration for a license, each feature **must** have the total amount of licenses and a limit of how many licenses can be booked by LM.
_Example:_ `features: {"example_feature": {"total": 100, "limit": 90}}`
If all licenses can be booked, the limit must match the total value. If the limit is lower than the total, bookings won't be created once the used amount of licenses reaches the limit. In this case, the job will be put back in the queue.

#### Why
This feature was requested to ensure that a configurable amount of licenses will be available to be used by desktop applications. Without the limit, the jobs running in the cluster can use all the licenses, which prevents the desktop application from running.

#### Additional Information
The configuration rows in the backend will need to be updated to use the new format for the feature field. For those that don't need this functionality, the limit == total. We'll need a list of each license that require the limit, and also which value should be used as the limit.

`Task`: https://app.clickup.com/t/18022949/ARMADA-815

---

#### Peer Review

### DIFF
--- a/agent/CHANGELOG.rst
+++ b/agent/CHANGELOG.rst
@@ -7,6 +7,10 @@ This file keeps track of all notable changes to license-manager-agent
 Unreleased
 ----------
 
+2.2.16-alpha.0 -- 2022-11-07
+----------------------------
+* Updated configuration row to use new feature quantities format
+
 2.2.15 -- 2022-10-26
 --------------------
 * Updated route to fetch licenses' configuration to use cluster specific route
--- a/agent/CHANGELOG.rst
+++ b/agent/CHANGELOG.rst
@@ -6,9 +6,6 @@ This file keeps track of all notable changes to license-manager-agent
 
 Unreleased
 ----------
-
-2.2.16-alpha.0 -- 2022-11-07
-----------------------------
 * Updated configuration row to use new feature quantities format
 
 2.2.15 -- 2022-10-26
--- a/agent/lm_agent/reconciliation.py
+++ b/agent/lm_agent/reconciliation.py
@@ -159,9 +159,10 @@ async def reconcile():
         config_id = await get_config_id_from_backend(product_feature)
         minimum_value = 0
         server_type = ""
+        (_, feature) = product_feature.split('.')
         for config in configs:
             if config.id == config_id:
-                minimum_value = config.features[product_feature.split(".")[1]]["total"]
+                minimum_value = config.features[feature]["total"]
                 server_type = config.license_server_type
                 break
         slurm_used = await get_tokens_for_license(product_feature + "@" + server_type, "Used")
--- a/agent/lm_agent/reconciliation.py
+++ b/agent/lm_agent/reconciliation.py
@@ -161,7 +161,7 @@ async def reconcile():
         server_type = ""
         for config in configs:
             if config.id == config_id:
-                minimum_value = config.features[product_feature.split(".")[1]]
+                minimum_value = config.features[product_feature.split(".")[1]]["total"]
                 server_type = config.license_server_type
                 break
         slurm_used = await get_tokens_for_license(product_feature + "@" + server_type, "Used")
--- a/agent/tests/test_backend_utils.py
+++ b/agent/tests/test_backend_utils.py
@@ -198,7 +198,12 @@ async def test_get_config_from_backend__omits_invalid_config_rows(
                 # Valid config row
                 dict(
                     product="SomeProduct",
-                    features={"A": 1, "list": 2, "of": 3, "features": 4},
+                    features={
+                        "A": {"total": 1, "limit": 1},
+                        "list": {"total": 2, "limit": 2},
+                        "of": {"total": 3, "limit": 3},
+                        "features": {"total": 4, "limit": 4},
+                    },
                     license_servers=["A", "list", "of", "license", "servers"],
                     license_server_type="O-Negative",
                     grace_time=13,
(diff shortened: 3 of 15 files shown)


backend/lm_backend/api/booking.py
```diff
@@ -96,9 +105,30 @@ async def get_bookings_job(job_id: str):
     return [BookingRow.parse_obj(x) for x in fetched]
 
 
+async def _get_limit_for_booking_feature(product_feature: str):
```
fschuch: Could you annotate the returning type?

backend/tests/api/test_booking.py
```diff
@@ -580,3 +610,19 @@ async def test_is_booking_available(
...
     assert await booking._is_booking_available(booking_row) is True
+
+
+@mark.asyncio
+@database.transaction(force_rollback=True)
+async def test_get_limit_for_booking_feature(
+    some_config_rows,
+    some_licenses,
+    insert_objects,
+):
```
fschuch, line 621: Docstring is missing

fschuch (changes requested):

backend/lm_backend/api/booking.py
julianaklulo: Sure, I added it in the latest commit.

backend/tests/api/test_booking.py
julianaklulo: Thanks, I added it in the latest commit.

agent/CHANGELOG.rst
```diff
@@ -7,6 +7,10 @@ This file keeps track of all notable changes to license-manager-agent
 Unreleased
 ----------
 
+2.2.16-alpha.0 -- 2022-11-07
```
dusktreader: Please put changes only under Major, Minor, and Patch versions in the CHANGELOG.

agent/lm_agent/reconciliation.py
```diff
@@ -161,7 +161,7 @@ async def reconcile():
         server_type = ""
         for config in configs:
             if config.id == config_id:
-                minimum_value = config.features[product_feature.split(".")[1]]
+                minimum_value = config.features[product_feature.split(".")[1]]["total"]
```
dusktreader: It would be good to pull the split out into an assigned variable. This would avoid having to do it in the loop over and over and it would also make it a bit more readable:
```python
(_,feature) = product_feature.split(".")
for config in configs:
    if config.id == config_id:
        minimum_value = config.features[feature]["total"]
```
```

File: agent/tests/test_backend_utils.py [hunk 5]
```diff
@@ -198,7 +198,12 @@ async def test_get_config_from_backend__omits_invalid_config_rows(
                 # Valid config row
                 dict(
                     product="SomeProduct",
-                    features={"A": 1, "list": 2, "of": 3, "features": 4},
+                    features={
+                        "A": {"total": 1, "limit": 1},
```
dusktreader, line 202: do we need to handle cases where the config row might not have a limit?

backend/CHANGELOG.rst
```diff
@@ -7,6 +7,10 @@ This file keeps track of all notable changes to license-manager-backend
 Unreleased
 ----------
 
+2.2.16-alpha.0 -- 2022-11-07
```
dusktreader: Same comment as above. Only Major, Minor, and Patch versions in the CHANGELOG, please.

backend/lm_backend/api/booking.py
```diff
@@ -96,9 +105,30 @@ async def get_bookings_job(job_id: str):
...
+    If the limit is the same as the total amount of licenses, all licenses can be booked.
+    """
+    product, feature = product_feature.split(".")
+
+    # Use product name to fetch config item from database
+    query = config_table.select().where(config_table.c.product == product)
+    fetched = await database.fetch_one(query)
+    config_row = ConfigurationRow.parse_obj(fetched)
+    config_item = ConfigurationItem(
+        **config_row.dict(exclude={"features"}), features=literal_eval(config_row.features)
```
dusktreader, line 120: Did you look at adding a nested schema for the features?

dusktreader: Thinking more about this and looking at the code, I think it would be very helpful to constraint the serialization/deserialization of the features to only insertion and extraction from the database. You could set it up so that throughout the rest of the code, you are dealing with a nested pydantic object, and it's only when you actually go to save it into the database that it's converted to a string.

On that note, we also really need to figure out how to get JSON and JSONB working in our apps. It's a perfect fit for this use-case.

lm-cli/CHANGELOG.rst
```diff
@@ -7,6 +7,10 @@ This file keeps track of all notable changes to License Manager CLI.
 Unreleased
 ----------
 
+2.2.16-alpha.0 -- 2022-11-07
```
dusktreader: Only Major, Minor, Patch in the Changelog...

dusktreader (changes requested):

agent/CHANGELOG.rst
julianaklulo: I wanted to publish this prerelease to test it in QA, but I'll remove it from all projects and test locally.

agent/lm_agent/reconciliation.py
julianaklulo: Thanks for the suggestion, I addressed in the latest commit.

agent/tests/test_backend_utils.py
julianaklulo: For now, all the config rows will have the total and the limit. If the license doesn't need the limit, its value will be the same as the total.

I suggest we take some time in the future to reestructure the data in the backend. I'm sure there's a better way of handling the features without having it as a JSON object serialized in a string field.

backend/lm_backend/api/booking.py
julianaklulo: I didn't. I followed this pattern of excluding the features string and evaluating it to a dict because it's the way it's being handled in all the other parts of the code.
But again, we should take some time to look into refactoring it.

fschuch (approved):

dusktreader (approved):

PR #187, opened 2023-01-13
[ASP-2214] Move utils module to main folder

julianaklulo:
#### What
Move utils model to main folder.
It was in the `server_interfaces` folder before because it was only used by the license server interfaces.

#### Why
This file will be used by the reservation interface as well.

`Task`: https://jira.scania.com/browse/ASP-2214

---

#### Peer Review

### DIFF
--- a/agent/lm_agent/server_interfaces/lsdyna.py
+++ b/agent/lm_agent/server_interfaces/lsdyna.py
@@ -5,7 +5,7 @@ from lm_agent.config import settings
 from lm_agent.exceptions import LicenseManagerBadServerOutput
 from lm_agent.parsing import lsdyna
 from lm_agent.server_interfaces.license_server_interface import LicenseReportItem, LicenseServerInterface
-from lm_agent.server_interfaces.utils import run_command
+from lm_agent.utils import run_command
 
 
 class LSDynaLicenseServer(LicenseServerInterface):
@@ -37,7 +37,7 @@ class LSDynaLicenseServer(LicenseServerInterface):
             output = await run_command(cmd)
 
             # try the next server if the previous didn't return the expected data
-            if output is None:
+            if not output:
                 continue
             return output
 
--- a/agent/lm_agent/utils.py
+++ b/agent/lm_agent/utils.py
@@ -7,13 +7,14 @@ from lm_agent.exceptions import CommandFailedToExecute
 from lm_agent.logs import logger
 
 
-async def run_command(command_line: str) -> str:
+async def run_command(*command_line_parts: str) -> str:
     """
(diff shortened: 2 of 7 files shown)


agent/lm_agent/server_interfaces/lsdyna.py
```diff
@@ -37,7 +37,7 @@ async def get_output_from_server(self):
             output = await run_command(cmd)
 
             # try the next server if the previous didn't return the expected data
-            if output is None:
+            if not output:
```
dusktreader, line 40: I wonder if getting the output via `run_command` should be put in the base-class or a mixin class so we don't implement it in each interface?

agent/lm_agent/utils.py
```diff
@@ -17,6 +18,10 @@ async def run_command(command_line: str):
...
 
     if proc.returncode != 0:
-        logger.error(f"Error: {output} | Return Code: {proc.returncode}")
-        return None
+        logger.error(
+            f"Command {command_line} failed!",
+            f"Error: {output}",
+            f"Return code: {proc.returncode}",
+        )
+        return False
```
dusktreader: I'm wondering if it would be better to raise an exception here instead. It would make it clearer when there is an error condition. This is especially true since the docstring doesn't explain that False indicates failure. It would be nicer if you could trust that any output from a function is valid and an error condition always results in an exception.

There could be some confusion if the output is empty, for example, but the command does not fail. In this case, the return value would still be falsey and fail the tests you had earlier, but there would be no log record indicating what the error was.

dusktreader (changes requested):

agent/lm_agent/server_interfaces/lsdyna.py
julianaklulo: The `run_command` method is going to be used by the reservation interface as well, that's why I moved it to a general `utils` module. Each server interface is calling the method from the `utils` module, not reimplementing it.

Do you think it's still worth adding the method in the base class, considering the reservation interface will need it too? If so, it would be outside the scope of this task.

agent/lm_agent/utils.py
julianaklulo: Yes, I refactored the function as you suggested:

- Returns the output as string if the command succeeds.
- Raises CommandFailedToExecute exception if return code is not zero.

agent/lm_agent/server_interfaces/lsdyna.py
dusktreader: No. I didn't realize it was being used for other cases as well. Disregard!

dusktreader (approved):

PR #188, opened 2023-01-23
[ASP-2214]  Create Slurm reservation interface

julianaklulo:
#### What
Create CRUD methods to interact with Slurm reservations.

#### Why
To be used by License Manager to improve license counter syncronization in the cluster.

`Task`: https://jira.scania.com/browse/ASP-2579

---

#### Peer Review

### DIFF
--- a/agent/lm_agent/config.py
+++ b/agent/lm_agent/config.py
@@ -69,6 +69,9 @@ class Settings(BaseSettings):
     # reservation name for reconciliation
     RESERVATION_IDENTIFIER: str = "license-manager-reservation"
 
+    # License Manager user name to create reservations
+    LM_USER: str = "license-manager"
+
     # debug mode turns on certain dangerous operations
     DEBUG: bool = False
 
--- a/agent/lm_agent/config.py
+++ b/agent/lm_agent/config.py
@@ -66,6 +66,9 @@ class Settings(BaseSettings):
     # path to the binary for olixtool (needed for OLicense licenses)
     OLIXTOOL_PATH: Path = _DEFAULT_BIN_PATH
 
+    # reservation name for reconciliation
+    RESERVATION_IDENTIFIER: str = "license-manager-reservation"
+
     # debug mode turns on certain dangerous operations
     DEBUG: bool = False
 
--- a/agent/lm_agent/workload_managers/slurm/reservations.py
+++ b/agent/lm_agent/workload_managers/slurm/reservations.py
@@ -30,14 +30,13 @@ async def scontrol_create_reservation(licenses: str, duration: str) -> bool:
     ]
 
     logger.debug(f"#### Creating reservation for {licenses} with duration {duration} ####")
-    reservation_output = await run_command(shlex.join(cmd))
+    reservation_output = await run_command(*cmd)
 
     logger_error_message = f"#### Failed to create reservation {settings.RESERVATION_IDENTIFIER} ####"
     if not reservation_output:
         logger.error(logger_error_message)
         return False
-    if isinstance(reservation_output, str):
-        if f"Reservation created: {settings.RESERVATION_IDENTIFIER}" not in reservation_output:
+    if f"Reservation created: {settings.RESERVATION_IDENTIFIER}" not in reservation_output:
             logger.error(logger_error_message)
             return False
 
@@ -59,14 +58,13 @@ async def scontrol_show_reservation() -> Union[str, bool]:
     ]
 
     logger.debug(f"#### Getting reservation {settings.RESERVATION_IDENTIFIER} ####")
-    reservation_output = await run_command(shlex.join(cmd))
+    reservation_output = await run_command(*cmd)
 
     logger_error_message = f"#### Failed to read reservation {settings.RESERVATION_IDENTIFIER} ####"
     if not reservation_output:
         logger.error(logger_error_message)
         return False
-    if isinstance(reservation_output, str):
-        if f"Reservation {settings.RESERVATION_IDENTIFIER} not found" in reservation_output:
+    if f"Reservation {settings.RESERVATION_IDENTIFIER} not found" in reservation_output:
             logger.error(logger_error_message)
             return False
 
@@ -93,14 +91,13 @@ async def scontrol_update_reservation(licenses: str, duration: str) -> bool:
     ]
 
     logger.debug(f"#### Updating reservation {settings.RESERVATION_IDENTIFIER} ####")
-    reservation_output = await run_command(shlex.join(cmd))
+    reservation_output = await run_command(*cmd)
 
     logger_error_message = f"#### Failed to update reservation {settings.RESERVATION_IDENTIFIER} ####"
     if not reservation_output:
         logger.error(logger_error_message)
         return False
-    if isinstance(reservation_output, str):
-        if "Reservation updated." not in reservation_output:
+    if "Reservation updated." not in reservation_output:
             logger.error(logger_error_message)
             return False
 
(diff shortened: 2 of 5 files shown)


agent/lm_agent/utils.py
```diff
@@ -17,6 +18,10 @@ async def run_command(command_line: str):
     output = str(stdout, encoding=ENCODING)
 
     if proc.returncode != 0:
-        logger.error(f"Error: {output} | Return Code: {proc.returncode}")
-        return None
+        logger.error(
+            f"Command {command_line} failed!",
+            f"Error: {output}",
+            f"Return code: {proc.returncode}",
+        )
```
matheushent: I just tested this and I couldn't get the three messages printed. As far as I know, only the first argument is the message to be printed, right? [Reference](https://docs.python.org/3.10/library/logging.html#logging.error).

Also, @dusktreader once recommended that this kind of log should be done only in one line instead of multiple and I agree, it is better indeed. Do you mind changing it?

agent/lm_agent/workload_managers/slurm/reservations.py
```diff
@@ -0,0 +1,132 @@
...
+
+    logger.debug(f"#### Creating reservation for {licenses} with duration {duration} ####")
+    reservation_output = await run_command(shlex.join(cmd))
+
+    logger_error_message = f"#### Failed to create reservation {settings.RESERVATION_IDENTIFIER} ####"
+    if not reservation_output:
+        logger.error(logger_error_message)
+        return False
+    if isinstance(reservation_output, str):
+        if f"Reservation created: {settings.RESERVATION_IDENTIFIER}" not in reservation_output:
```
matheushent: this is bad. I had problems with it to debug the lm agent charm. Is there a way to check the return code?

agent/lm_agent/config.py
```diff
@@ -66,6 +66,12 @@ class Settings(BaseSettings):
     # path to the binary for olixtool (needed for OLicense licenses)
     OLIXTOOL_PATH: Path = _DEFAULT_BIN_PATH
 
+    # reservation name for reconciliation
+    RESERVATION_IDENTIFIER: str = "license-manager-reservation"
+
+    # License Manager user name to create reservations
+    LM_USER: str = "license-manager"
```
matheushent, line 73: I have a conceptual question about it: shouldn't the reservation be made on behalf of the user who requested the reservation, i.e. someone real?

agent/lm_agent/utils.py
dusktreader: Yes, the reason is that you want log messages that are part of a single message to appear as one entry in log aggregators and such. In this case, I think you could just join each part with a newline.

agent/lm_agent/workload_managers/slurm/reservations.py
dusktreader: In the other PR, I commented on this. It would be good if the `run_command` function always returned a string when the command succeeded and raised an exception if the command failed. This would eliminate this extra step of checking return type.

agent/lm_agent/workload_managers/slurm/reservations.py
```diff
@@ -0,0 +1,132 @@
...
+        SCONTROL_PATH,
+        "update",
+        "reservation",
+        f"ReservationName={settings.RESERVATION_IDENTIFIER}",
+        f"duration={duration}",
+        f"licenses={licenses}",
+    ]
+
+    logger.debug(f"#### Updating reservation {settings.RESERVATION_IDENTIFIER} ####")
+    reservation_output = await run_command(shlex.join(cmd))
```
dusktreader: Since this pattern is being repeated all over the place, maybe it would be nice to put the join in the method itself?

```
async def run_command(*parts):
    command = shlex.join(parts)
    ...
```

dusktreader (changes requested):

agent/lm_agent/config.py
julianaklulo: No, the reservation is made by license manager to block licenses that are already in use in other environments.
We'll provision an operator level user to run the agent that will be able to manage the reservation.
This reservation is not to be used by any users, it's only to sync the counters for licenses in use.

agent/lm_agent/utils.py
julianaklulo: You're right, I fixed it in #189.

agent/lm_agent/workload_managers/slurm/reservations.py
julianaklulo: I updated the `run_command` function to to as @dusktreader suggested.
I'll update the logic in the methods taking this into consideration.

agent/lm_agent/workload_managers/slurm/reservations.py
julianaklulo: @dusktreader Great idea, I updated it in #187 and will change the code here to take it into consideration.

agent/lm_agent/workload_managers/slurm/reservations.py
julianaklulo: @matheushent now the `run_command` function will raise an exception if the return code != 0.
This made possible to remove the output checking here.
Is it better now?

agent/lm_agent/workload_managers/slurm/reservations.py
```diff
@@ -0,0 +1,116 @@
...
+        "show",
+        "reservation",
+        settings.RESERVATION_IDENTIFIER,
+    ]
+
+    logger.debug(f"#### Getting reservation {settings.RESERVATION_IDENTIFIER} ####")
+    try:
+        reservation_output = await run_command(*cmd)
+    except CommandFailedToExecute:
+        logger.debug(f"#### Failed to read reservation {settings.RESERVATION_IDENTIFIER} ####")
```
matheushent:
```suggestion
        logger.error(f"#### Failed to read reservation {settings.RESERVATION_IDENTIFIER} ####")
```

agent/lm_agent/workload_managers/slurm/reservations.py
```diff
@@ -0,0 +1,116 @@
...
+        f"duration={duration}",
+        "flags=license_only",
+        f"licenses={licenses}",
+    ]
+
+    logger.debug(f"#### Creating reservation for {licenses} with duration {duration} ####")
+    try:
+        await run_command(*cmd)
+    except CommandFailedToExecute:
+        logger.debug(f"#### Failed to create reservation {settings.RESERVATION_IDENTIFIER} ####")
```
matheushent:
```suggestion
        logger.error(f"#### Failed to create reservation {settings.RESERVATION_IDENTIFIER} ####")
```

agent/lm_agent/workload_managers/slurm/reservations.py
julianaklulo: Thank you!

agent/lm_agent/workload_managers/slurm/reservations.py
julianaklulo: Thank you!

agent/lm_agent/workload_managers/slurm/reservations.py
matheushent: @julianaklulo I'm a little bit confused because I see the try-except blocks but I couldn't find the change in the function in this PR. I also see some checks by strings ([here](https://github.com/omnivector-solutions/license-manager/pull/188/commits/15aefb998d938dd0fec1b504cfd47ad11e3dca7a#diff-b969ae58520282e57a72883f59250a6cc46181fc9cbe4abd03c08c8faa7513d1R100) for example)

julianaklulo: I removed all checks by strings in [ec65243](https://github.com/omnivector-solutions/license-manager/pull/188/commits/ec65243406e16a7439711b0127eec891c11deb9a). Please look at the files changed to see the new version of the code.

julianaklulo: The modification in the `run_command` function to raise the exception is already in main [32d35f7](https://github.com/omnivector-solutions/license-manager/commit/32d35f7f7b7111a8e1ce10b682480b03a5317641), it was done in #187 that had been opened before this PR and is already merged.

matheushent (approved):

dusktreader (approved):

PR #201, opened 2023-02-16
[ASP-2544]  Update reconcile to use Slurm reservation

julianaklulo:
#### What
Update the reconcile method to create a reservation to block licenses in use in the license server and/or other clusters instead of changing the total of the license.

#### Why
To keep the counters in sync with the actual usage.

`Task`: https://jira.scania.com/browse/ASP-2544
---

#### Peer Review


agent/lm_agent/reconciliation.py
```diff
@@ -144,55 +145,146 @@ async def filter_cluster_update_licenses(licenses_to_update: List) -> List:
...
     """Generate the report and reconcile the license feature token usage."""
+    logger.debug("Starting reconciliation")
+    # Delete bookings for jobs that reached the grace time
+    logger.debug("Cleaning bookings by grace time")
     await clean_booked_grace_time()
-    r = await update_report()
+    logger.debug("Bookings cleaned by grace time")
+
+    # Generate report and update the backend
+    logger.debug("Reconciliating licenses in the backend")
```
dusktreader:
```suggestion
    logger.debug("Reconciling licenses in the backend")
```

agent/lm_agent/reconciliation.py
```diff
@@ -144,55 +145,146 @@ async def filter_cluster_update_licenses(licenses_to_update: List) -> List:
...
+    If the reservation cannot be updated, delete it and create a new one.
+    """
+    reservation = await scontrol_show_reservation()
+
+    if reservation:
+        updated = await scontrol_update_reservation(reservation_data, "30:00")
+        if not updated:
+            deleted = await scontrol_delete_reservation()
+            if not deleted:
+                raise LicenseManagerReservationFailure("Could not update or delete reservation.")
```
dusktreader: You could use the `require_condition()` helper here for some syntactical sugar.

agent/lm_agent/reconciliation.py
```diff
@@ -144,55 +145,146 @@ async def filter_cluster_update_licenses(licenses_to_update: List) -> List:
...
-        update_resource = await sacctmgr_modify_resource(product, feature, new_quantity)
-        if update_resource:
-            logger.info("Slurmdbd updated successfully.")
-        else:
-            logger.info("Slurmdbd update unsuccessful.")
-    return r
+        server_used = license_data["license_used"]
+        cluster_name = await get_cluster_name()
+        cluster_booking_sum = await get_booking_sum_for_cluster(cluster_name)
+        other_cluster_booking_sum = await get_booking_sum_for_other_clusters(cluster_name)
```
dusktreader: There is a lot of inefficiency here in calling these two methods consecutively. The functions are exactly the same except for the comparison operator that is used in the conditional sum. You're making the same API call for both and iterating over results for both. Basically, the work is doubled without need.

I think these need to be collapsed into one method that counts _both_ sums in one pass.

agent/lm_agent/reconciliation.py
```diff
@@ -144,55 +145,146 @@ async def filter_cluster_update_licenses(licenses_to_update: List) -> List:
...
+        config = await backend_client.get(f"/lm/api/v1/config/{config_id}")
+        config = config.json()
+
+        license_server_type = config["license_server_type"]
+        # Use feature name to get total and limit from feature data in the license config
+        try:
+            # Get total from new feature format
+            total = config["features"][feature].get("total")
+            LicenseManagerFeatureConfigurationIncorrect.require_condition(
+                total,
```
dusktreader, line 235: this condition will be raised if the total is 0. I'm assuming that it shouldn't ever be zero, but the message could be confusing in that case.

agent/lm_agent/reconciliation.py
```diff
@@ -144,55 +145,146 @@ async def filter_cluster_update_licenses(licenses_to_update: List) -> List:
...
+
+        try:
+            # Get limit from new feature format. If not specified, use the total as the limit
+            limit = config["features"][feature].get("limit", total)
+        except AttributeError:
+            # Fallback to use the total as the limit for the old feature format
+            limit = total
+
+        # Get license usage from the cluster
+        slurm_used = await get_tokens_for_license(product_feature + "@" + license_server_type, "Used")
```
dusktreader: Probably best to use an f-string here if only for consistency (see line 281).

dusktreader (changes requested):
Nice work! I have some comments for you.

agent/lm_agent/reconciliation.py
julianaklulo: I updated it in the latest commit.

agent/lm_agent/reconciliation.py
julianaklulo: Totally, I updated it to be only one function that returns the booking sum for all clusters. The reconcile extracts the needed information from the result doing the logic needed here.

agent/lm_agent/reconciliation.py
julianaklulo: The exception should be raised if the key "total" doesn't exist in the feature's dict.
I updated the .get("total") to return 0 in that case, which will make this the expected condition for the exception validation.

agent/lm_agent/reconciliation.py
julianaklulo: Yes, this was a left over from the previous version of the code.

dusktreader (approved):

PR #202, opened 2023-02-22
Fix run command issue

julianaklulo:
#### What
Fix issue with trailing quote in `shlex.join()` function.

#### Why
It was making the reconcile process fail.

`Task`: 

---

#### Peer Review


agent/lm_agent/utils.py
```diff
@@ -14,7 +14,7 @@ async def run_command(*command_line_parts: str) -> str:
     Raises CommandFailedToExecute exception if return code is not zero.
     """
 
-    command_line = shlex.join(command_line_parts)
+    command_line = shlex.join(command_line_parts).strip("'")
```
dusktreader: I did a little digging on this solution. I don't think this is the right way to approach it. Consider a command argv passed to `run_command` such as `['echo', 'hello world']`. The result of joining with shlex is `"echo 'hello world'"`. Stripping the single quote would result in `echo 'hello world`, which is an invalid command.

If we look at how the flexlm module is calling `run_command`, it is doing it like this:
```python
feature_cmd = f"{cmd} {feature}"
output = await run_command(feature_cmd)
```
The problem that we are seeing is that we are passing a single argument that is an string that has not been split into an argv array. Thus, calling join on it is adding that extra quote:

```
>>> shlex.join(["foo bar"])
"'foo bar'"
```

I think the correct fix will be to instead pass the `cmd` and `feature` variables as separate arguments to `run_command()`:

```python
output = await run_command(cmd, feature)
```

I think the `strip()` fix here is a quick hack that will cost us in the long run.

dusktreader (changes requested):

agent/lm_agent/utils.py
julianaklulo: @dusktreader I ran some tests in QA and just `output = await run_command(cmd, feature)` didn't work, because `shlex.join()` expects a list.
I changed the command parts in each server interface to send a list to `run_command`, and also changed `run_command `to accept a list as its argument and it worked.
It seems to be the best approach for a long term solution.

dusktreader (approved):

PR #203, opened 2023-02-28
[ASP-2858] Fix run command issue

julianaklulo:
#### What
Port the fix for the run command issue to the main branch (originally made for release 2.2.22 in #202)

#### Why
The fix was applied only in the branch that needed the fix at the moment.

`Task`: https://jira.scania.com/browse/ASP-2858

---

#### Peer Review

### DIFF
--- a/agent/tests/server_interfaces/test_flexlm.py
+++ b/agent/tests/server_interfaces/test_flexlm.py
@@ -1,5 +1,6 @@
 """Test the FlexLM license server interface."""
 
+import shlex
 from unittest import mock
 
 from pytest import fixture, mark, raises
@@ -21,7 +22,7 @@ def test_get_flexlm_commands_list(flexlm_server: FlexLMLicenseServer):
     Do the commands for invoking the license server have the correct data?
     """
     commands_list = flexlm_server.get_commands_list()
-    assert commands_list == [f"{settings.LMUTIL_PATH} lmstat -c 2345@127.0.0.1 -f"]
+    assert shlex.join(commands_list[0]) == f"{settings.LMUTIL_PATH} lmstat -c 2345@127.0.0.1 -f"
 
 
 @mark.asyncio
--- a/agent/tests/server_interfaces/test_flexlm.py
+++ b/agent/tests/server_interfaces/test_flexlm.py
@@ -22,7 +22,15 @@ def test_get_flexlm_commands_list(flexlm_server: FlexLMLicenseServer):
     Do the commands for invoking the license server have the correct data?
     """
     commands_list = flexlm_server.get_commands_list()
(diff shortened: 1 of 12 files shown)


agent/tests/server_interfaces/test_flexlm.py
```diff
@@ -21,7 +22,7 @@ def test_get_flexlm_commands_list(flexlm_server: FlexLMLicenseServer):
     Do the commands for invoking the license server have the correct data?
     """
     commands_list = flexlm_server.get_commands_list()
-    assert commands_list == [f"{settings.LMUTIL_PATH} lmstat -c 2345@127.0.0.1 -f"]
+    assert shlex.join(commands_list[0]) == f"{settings.LMUTIL_PATH} lmstat -c 2345@127.0.0.1 -f"
```
matheushent: not that will be much different in this case, but I feel strange parsing the result of a function we want to test and then check the result. Who knows what happens during this parse. I believe this assertion would be better placed like:
```suggestion
    assert commands_list == [settings.LMUTIL_PATH, "lmstat", "-c", "2345@127.0.0.1", "-f"]
```

The reason I do think this is better is because we indeed verify the behaviour of the function but it is not needed any extra helper.

julianaklulo: I improved the tests, is it better now?

agent/tests/server_interfaces/test_olicense.py
```diff
@@ -20,7 +21,13 @@ def test_get_olicense_commands_list(olicense_server: OLicenseLicenseServer):
...
     """
     commands_list = olicense_server.get_commands_list()
-    assert commands_list == [f"{settings.OLIXTOOL_PATH} -sv 127.0.0.1:2345"]
+    assert commands_list == [
+        [
+            f"{settings.OLIXTOOL_PATH}",
+            "-sv",
+            "127.0.0.1:2345",
+        ]
+    ]
```
matheushent, line 30: so, I'm confused why the return for the `get_commands_list` is a list of lists. I mean, each license server has only one command (I noticed by the tests), right? If so, why to return a list of lists?

matheushent: was it designed in case there are more than one license server per license?

agent/tests/server_interfaces/test_flexlm.py
matheushent: yes, thank you

agent/tests/server_interfaces/test_olicense.py
julianaklulo: Each license configuration can have `n` license servers, so this function needs to generate the command for each entry.

julianaklulo: Some license servers have redundant mirrors, in case one of them fails we use the next one and so on.

dusktreader (approved):

PR #205, opened 2023-03-01
[ASP-2544] Change access token permission

julianaklulo:
#### What
Change the cached access token creation to create one file for each user running the reconcile process.
Also improve log messages regarding token read/write operations.

#### Why
The agent runs as user `license-manager` (not `root` anymore), which is causing issues with the access token file owned by user `slurm`.
To solve this, each user will create its own access token in the cache directory, with permission 600 (read/write only for the owner). This also addresses security concerns regarding sharing a token file with multiple users.

`Task`: https://jira.scania.com/browse/ASP-2544

---

#### Peer Review

### DIFF
--- a/agent/lm_agent/backend_utils.py
+++ b/agent/lm_agent/backend_utils.py
@@ -1,7 +1,7 @@
 """
 Provide utilities that communicate with the backend.
 """
-import shutil
+import getpass
 import typing
 
 import httpx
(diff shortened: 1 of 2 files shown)


agent/lm_agent/backend_utils.py
```diff
@@ -54,9 +55,9 @@ def _write_token_to_cache(token: str):
 
     token_path = settings.CACHE_DIR / TOKEN_FILE_NAME
     try:
-        token_path.touch(mode=0o600, exist_ok=True)
+        token_path.touch(mode=0o660, exist_ok=True)
```
dusktreader: As we discussed in Slack, I think we will need to keep separate files for the different users. We need to make sure that token files are stored with at least 0o600 permissions.

Additionally, I suggest using `chmod` after the `write_text()` call every time instead of using `touch()`.

dusktreader (changes requested):

agent/lm_agent/backend_utils.py
julianaklulo: I updated the code to create one file for each user, with permission 600. Also removed the `touch()` function to use `chmod` instead.

dusktreader (approved):

PR #206, opened 2023-03-02
Change aws-psycopg2 to psycopg2

julianaklulo:
#### What
Change the dependency aws-psycopg2 to psycopg2.

#### Why
The package aws-psycopg2 is breaking the GitHub Action with Poetry 1.4.0.


matheushent (approved):
As talked in private conversations, let's just make sure it works as expected by using the Vantage CLI.

PR #220, opened 2023-05-10
[ASP-2210] Harden License Manage data model

julianaklulo:
#### What
Rewrite the API to use a normalized database.

#### Why
To improve the usability of the API.

`Task`: https://jira.scania.com/browse/ASP-2210

---

#### Peer Review


backend/Dockerfile
```diff
@@ -5,14 +5,14 @@ WORKDIR /app
...
-    POETRY_HOME=/opt/poetry POETRY_VERSION=1.1.13 python && \
+    POETRY_HOME=/opt/poetry POETRY_VERSION=1.4.0 python && \
     cd /usr/local/bin && \
     ln -s /opt/poetry/bin/poetry && \
     poetry config virtualenvs.create false
 
 COPY ./pyproject.toml ./poetry.lock* /app/
 
-RUN poetry install --no-root --no-dev
+RUN poetry install --no-root
```
dusktreader: Why do we need the dev dependencies in the Docker images?

If we need this for our CI pipeline, then we should create a separate version (Dockerfile.ci) that includes these and that is used by the docker-compose.yaml used in the CI. Our docker images in prod should not include dev dependencies.

backend/alembic/env.py
```diff
@@ -1,11 +1,12 @@
...
+
 from logging.config import fileConfig
 
 from alembic import context
 from sqlalchemy import engine_from_config, pool
 from sqlalchemy_utils import ScalarListType
 
-from lm_backend.config import settings
-from lm_backend.table_schemas import metadata as lm_backend_metadata
+sys.path = ['', '..'] + sys.path[1:]
```
dusktreader: Why do we have this here?

backend/alembic/versions/0003--implement_new_data_struture_schema.py
```diff
@@ -0,0 +1,195 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "594b52f2c088"
+down_revision = "5308f26e8b60"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
dusktreader: These comments should be removed after the auto-generated migrations are reviewed by the dev.

backend/alembic/versions/0003--implement_new_data_struture_schema.py
```diff
@@ -0,0 +1,195 @@
...
+    op.drop_table("license")
+    op.drop_table("booking")
+    op.drop_index("ix_config_client_id", table_name="config")
+    op.drop_table("config")
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
```
dusktreader: I don't think we should have downgrade commands here. This will be a breaking change. We should not ever go back once we apply this.

backend/docker-compose.yaml
```diff
@@ -30,3 +30,52 @@ services:
...
+      context: .
+      dockerfile: ./Dockerfile
+    environment:
+      DATABASE_URL: postgresql://omnivector:rats@pgsql:5432/ratsdb
+      ARMASEC_DOMAIN: ${ARMASEC_DOMAIN}
+      ARMASEC_AUDIENCE: ${ARMASEC_AUDIENCE}
+      ARMASEC_DEBUG: ${ARMASEC_DEBUG}
+      LOG_LEVEL: ${LOG_LEVEL}
+    command: >
+      sh -c  "/usr/local/bin/pip install alembic==1.8.1 psycopg2==2.9.5 && alembic -c alembic/alembic.ini upgrade head"
```
dusktreader: Why doesn't this just use poetry? We should not be including hard-coded dependency specifications in a compose file.

backend/lm_backend/api/__init__.py
```diff
@@ -1,10 +1,20 @@
...
-api_v1.include_router(router_config, prefix="/config", tags=["Config"])
+api = APIRouter()
+api.include_router(router_clusters, prefix="/clusters", tags=["Cluster"])
+api.include_router(router_configurations, prefix="/configurations", tags=["Configuration"])
+api.include_router(router_license_servers, prefix="/license_servers", tags=["License Server"])
+api.include_router(router_products, prefix="/products", tags=["Product"])
+api.include_router(router_features, prefix="/features", tags=["Feature"])
+api.include_router(router_inventories, prefix="/inventories", tags=["Inventory"])
+api.include_router(router_jobs, prefix="/jobs", tags=["Job"])
+api.include_router(router_bookings, prefix="/bookings", tags=["Booking"])
```
dusktreader, line 20: Beautiful!!

backend/lm_backend/api/cruds/booking.py
```diff
@@ -0,0 +1,75 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from lm_backend.api.cruds.generic import CreateSchemaType, GenericCRUD, ModelType, UpdateSchemaType
+from lm_backend.api.models import Booking, Feature, Inventory
+from lm_backend.api.schemas import BookingCreateSchema
+
+
+class BookingCRUD(GenericCRUD):
+    def __init__(self, model: ModelType, create_schema: CreateSchemaType, update_schema: UpdateSchemaType):
+        super().__init__(model, create_schema, update_schema)
```
dusktreader: This `__init__()` method looks like it's unnecessary.

backend/lm_backend/api/cruds/booking.py
```diff
@@ -0,0 +1,75 @@
...
+
+    async def create(self, db_session: AsyncSession, obj=BookingCreateSchema) -> ModelType:
+        """Create a new booking.
+
+        Use a somewhat complex logic to check if a booking can be made and insert it in one query. In this
+        case, implement the `INSERT columns FROM select_query WHERE EXISTS (SELECT subquery)` SQL paradigm
+        using SQLAlchemy query API. A  booking will only be created if there are enough licenses available
+        to make the booking.
```
dusktreader, line 24: Good docstring!

backend/lm_backend/api/cruds/booking.py
```diff
@@ -0,0 +1,75 @@
...
+            .select_from(Feature)
+            .join(Booking, and_(Feature.id == Booking.feature_id, Booking.feature_id == obj.feature_id))
+            .join(Inventory, Feature.id == Inventory.feature_id)
+            .group_by(Feature.id)
+            .having(
+                func.sum(Booking.quantity)
+                + func.max(Inventory.used)
+                + obj.quantity
+                + func.max(Feature.reserved)
+                <= func.max(Inventory.total)
```
dusktreader: I think it would be good to write this inequality in the docstring in a clearer way if possible. This isn't terribly friendly to read in the sense of an inequality expression. It's easy to read as code, but I think it would be good if the core logic of this method was clearly expressed above.

backend/lm_backend/api/cruds/generic.py
```diff
@@ -0,0 +1,153 @@
...
+        self.create_schema = create_schema
+        self.update_schema = update_schema
+
+    async def create(self, db_session: AsyncSession, obj: CreateSchemaType) -> ModelType:
+        """Creates a new object in the database."""
+        db_obj = self.model(**obj.dict())
+        try:
+            async with db_session.begin():
+                db_session.add(db_obj)
+            await db_session.commit()
```
dusktreader: Is an explicit commit needed here? I thought you got an implicit commit when you use `begin()` as a context manager.

backend/lm_backend/api/cruds/generic.py
```diff
@@ -0,0 +1,153 @@
...
+                        setattr(db_obj, field, value)
+                await db_session.flush()
+            except Exception as e:
+                logger.error(e)
+                raise HTTPException(status_code=400, detail="Object could not be updated.")
+
+            await db_session.refresh(db_obj)
+        return db_obj
+
+    async def delete(self, db_session: AsyncSession, id: int) -> bool:
```
dusktreader: Why is this function signature returning a bool? There's no return expressions in here.

backend/lm_backend/api/models.py
```diff
@@ -0,0 +1,216 @@
+"""
+Database models for the database tables in the License Manager API.
+Tables:
+    - license_servers
```
dusktreader: Now that we have so many models, we really need to break this file up. I think maybe we should instead have a `models/` folder with a separate module for each one. Then the `__init__.py` module can import each one so that abstraction doesn't impact how other modules import them. What do you think?

backend/lm_backend/api/models.py
```diff
@@ -0,0 +1,216 @@
...
+    features = relationship(
+        "Feature", back_populates="configurations", lazy="selectin", cascade="all, delete-orphan"
+    )
+
+    searchable_fields = [name]
+    sortable_fields = [name, cluster_id]
+
+    def __repr__(self):
+        return (
+            f"Config(id={self.id}, "
```
dusktreader: The open paren after `Config` is never closed.

backend/lm_backend/api/routes/bookings.py
```diff
@@ -0,0 +1,71 @@
...
+from lm_backend.api.models import Booking
+from lm_backend.api.schemas import BookingCreateSchema, BookingSchema, BookingUpdateSchema
+from lm_backend.permissions import Permissions
+from lm_backend.security import guard
+from lm_backend.session import get_session
+
+router = APIRouter()
+
+
+crud_booking = BookingCRUD(Booking, BookingCreateSchema, BookingUpdateSchema)
```
dusktreader, line 16: I really like this approach for the crud endpoints. Wrapping the logic for mapping from user-supplied data to database entries and back is very nice. It makes the routers so clean for basic CRUD operations.

backend/lm_backend/api/routes/bookings.py
```diff
@@ -0,0 +1,71 @@
...
+    "/",
+    response_model=BookingSchema,
+    status_code=status.HTTP_201_CREATED,
+    dependencies=[Depends(guard.lockdown(Permissions.BOOKING_EDIT))],
+)
+async def create_booking(
+    booking: BookingCreateSchema,
+    db_session: AsyncSession = Depends(get_session),
+):
+    """Create a new booking."""
```
dusktreader, line 29: It might be nice to add some logging in the routes so that we can see request details when they come in.

backend/lm_backend/api/routes/clusters.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@router.delete(
+    "/{cluster_id}",
+    status_code=status.HTTP_200_OK,
+    dependencies=[Depends(guard.lockdown(Permissions.CLUSTER_EDIT))],
+)
+async def delete_cluster(cluster_id: int, db_session: AsyncSession = Depends(get_session)):
+    """
+    Delete a cluster from the database.
```
dusktreader, line 104: For multi-line docstrings there should be an empty line after the first one.

backend/lm_backend/api/routes/jobs.py
```diff
@@ -0,0 +1,117 @@
...
+@router.delete(
+    "/slurm_job_id/{slurm_job_id}/cluster/{cluster_id}",
+    status_code=status.HTTP_200_OK,
+    dependencies=[Depends(guard.lockdown(Permissions.JOB_EDIT))],
+)
+async def delete_job_by_slurm_id(
+    slurm_job_id: str, cluster_id: int, db_session: AsyncSession = Depends(get_session)
+):
+    """
+    Delete a job from the database and associated bookings.
```
dusktreader, line 82: Need a blank line after the first one for multi-line docstrings.

backend/lm_backend/api/routes/jobs.py
```diff
@@ -0,0 +1,117 @@
...
+    Since the slurm_job_id can be the same across clusters, we need the cluster_id to validate.
+    """
+    jobs = await crud_job.read_all(db_session=db_session, search=slurm_job_id)
+
+    for job in jobs:
+        if job.cluster_id == cluster_id:
+            await crud_job.delete(db_session=db_session, id=job.id)
+            return {"message": "Job deleted successfully"}
+
+    raise HTTPException(status_code=404, detail="The job doesn't exist in this cluster.")
```
dusktreader, line 94: Might be clearer to raise this exception if the `jobs` list is empty.

backend/lm_backend/api/routes/jobs.py
```diff
@@ -0,0 +1,117 @@
...
+    "/slurm_job_id/{slurm_job_id}/cluster/{cluster_id}",
+    status_code=status.HTTP_200_OK,
+    dependencies=[Depends(guard.lockdown(Permissions.JOB_VIEW))],
+)
+async def read_job_by_slurm_id(
+    slurm_job_id: str, cluster_id: int, db_session: AsyncSession = Depends(get_session)
+):
+    """
+    Read a job from the database and associated bookings.
+    Uses the slurm_job_id and the cluster_id to filter the job.
```
dusktreader, line 108: blank line after first in multiline docstring.

backend/lm_backend/api/routes/products.py
```diff
@@ -0,0 +1,89 @@
...
+
+@router.delete(
+    "/{product_id}",
+    status_code=status.HTTP_200_OK,
+    dependencies=[Depends(guard.lockdown(Permissions.PRODUCT_EDIT))],
+)
+async def delete_product(product_id: int, db_session: AsyncSession = Depends(get_session)):
+    """Delete a product from the database and associated features."""
+    await crud_product.delete(db_session=db_session, id=product_id)
+    return {"message": "Product deleted successfully"}
```
dusktreader: I wonder if you could make the GenericCRUD return a message like this with the name of the SQLAlchemy model in the string so you don't have to repeat this in every router delete.

backend/lm_backend/api/schemas.py
```diff
@@ -0,0 +1,363 @@
+"""
+Schemas for the License Manager API.
```
dusktreader: I wonder if it would be good to break the schemas up into a module for each of the different entities so they aren't all in one huge file.

backend/lm_backend/api/schemas.py
```diff
@@ -0,0 +1,363 @@
...
+
+class BaseUpdateSchema(BaseModel):
+    """
+    Base class for updating objects in the database.
+    """
+
+    pass
+
+
+class LicenseServerType(str, Enum):
```
dusktreader: This should probably go into a "constants.py" module or something similar since it isn't itself a schema.

backend/lm_backend/api/schemas.py
```diff
@@ -0,0 +1,363 @@
...
+    License server to be created in the database.
+    """
+
+    config_id: int
+    host: str
+    port: int
+    type: LicenseServerType
+
+    @validator("port")
+    def port_must_be_positive(cls, v):
```
dusktreader: I think you could use a constrained type from Pydantic for this: https://docs.pydantic.dev/latest/usage/types/#constrained-types

backend/lm_backend/database.py
```diff
@@ -1,31 +1,22 @@
 """
-Persistent data storage for the backend
+Persistent data storage for the API.
```
dusktreader, line 2: If we are going to rename "backend" to "API", it might make sense to change the subproject name as well? Maybe that's too much to take on right now.

backend/lm_backend/database.py
```diff
@@ -1,31 +1,22 @@
...
-    metadata.create_all(engine)
+Base = declarative_base()
 
 
 def render_sql(query) -> str:
     """
     Render a sqlalchemy query into a string for debugging.
     """
-    return query.compile(dialect=database._backend._dialect, compile_kwargs={"literal_binds": True})
+    return query.compile(compile_kwargs={"literal_binds": True})
```
dusktreader, line 19: Isn't the dialect needed to properly render the query args?

backend/tests/api/routes/conftest.py
```diff
@@ -0,0 +1,325 @@
...
+            "name": "converge_super",
+            "product_id": product_id,
+            "config_id": configuration_id,
+            "reserved": 0,
+        },
+    ]
+
+    inserted_features = await insert_objects(features_to_add, Feature)
+
+    # inventories_to_add = [
```
dusktreader: Should this just be removed?

backend/tests/api/routes/conftest.py
```diff
@@ -0,0 +1,325 @@
...
+            "name": "abaqus",
+            "product_id": product_id,
+            "config_id": configuration_id,
+            "reserved": 0,
+        },
+    ]
+
+    inserted_feature = await insert_objects(feature_to_add, Feature)
+
+    # inventory_to_add = [{
```
dusktreader: Remove?

backend/tests/api/routes/test_clusters.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+@mark.asyncio
+async def test_add_cluster__success(
+    backend_client: AsyncClient,
+    inject_security_header,
+    read_object,
+    clean_up_database,
+):
+    """
```
dusktreader: We should be consistent with docstrings on the test functions. I personally dont't think they are necessary. We should just be consistent. This docstring doesn't really add much.

backend/tests/api/routes/test_bookings.py
```diff
@@ -0,0 +1,154 @@
...
+    assert response.status_code == 200
+
+    response_booking = response.json()
+    assert response_booking["job_id"] == create_one_booking[0].job_id
+    assert response_booking["feature_id"] == create_one_booking[0].feature_id
+    assert response_booking["quantity"] == create_one_booking[0].quantity
+
+
+@mark.asyncio
+async def test_delete_booking__success(
```
dusktreader, line 176: I think we need a test case for failures on each endpoint where parameters can cause issues. So, get by id, delete, update, etc.

backend/tests/conftest.py
```diff
@@ -1,65 +1,181 @@
...
-    Make sure our database is empty at the end of each test
+    A test app to be used by the test backend_client.
     """
-    yield
 
-    count = await database.fetch_all("SELECT COUNT(*) FROM license")
-    assert count[0][0] == 0
+    from lm_backend.main import app as test_app
+
+    yield test_app
```
dusktreader, line 52: Why is this needed? Can't the test modules just import app from main.py themselves?

backend/tests/conftest.py
```diff
@@ -1,65 +1,181 @@
...
+    async with get_session() as sess:
+        await sess.execute(sqlalchemy.text(f"""DELETE FROM {models.Booking.__tablename__};"""))
+        await sess.execute(sqlalchemy.text(f"""DELETE FROM {models.Job.__tablename__};"""))
+        await sess.execute(sqlalchemy.text(f"""DELETE FROM {models.Inventory.__tablename__};"""))
+        await sess.execute(sqlalchemy.text(f"""DELETE FROM {models.Feature.__tablename__};"""))
+        await sess.execute(sqlalchemy.text(f"""DELETE FROM {models.Product.__tablename__};"""))
+        await sess.execute(sqlalchemy.text(f"""DELETE FROM {models.LicenseServer.__tablename__};"""))
+        await sess.execute(sqlalchemy.text(f"""DELETE FROM {models.Configuration.__tablename__};"""))
+        await sess.execute(sqlalchemy.text(f"""DELETE FROM {models.Cluster.__tablename__};"""))
+        await sess.commit()
```
dusktreader: This is going to make your tests very slow. The way to handle this is to make the session during unit tests to flush their objects instead of commit. The database should be rolled back after each test function finishes.

dusktreader (changes requested):

backend/Dockerfile
julianaklulo: I created a separate Dockerfile for the CI and reverted this change.

backend/alembic/env.py
julianaklulo: The [SOS API](https://github.com/omnivector-solutions/sos/blob/c169acae6b62a037e31d4ee9103ea682b561b7e6/sos-api/alembic/env.py#L10) uses it, @matheushent said it should prevent issues when running the migrations inside the container.
I tested running it without this line and it worked, so I removed it.

backend/alembic/versions/0003--implement_new_data_struture_schema.py
julianaklulo: I removed it.

backend/alembic/versions/0003--implement_new_data_struture_schema.py
julianaklulo: I removed the downgrade commands and the old migrations, leaving only the new migration.

backend/docker-compose.yaml
julianaklulo: This also came from the [SOS API](https://github.com/omnivector-solutions/sos/blob/main/sos-api/docker-compose.yaml#L84), I updated it to use poetry, since these dependencies are already in the pyproject.toml.

backend/lm_backend/api/cruds/booking.py
julianaklulo: I removed it.

backend/lm_backend/api/cruds/booking.py
julianaklulo: I updated the docstring to include:

> To determine if a booking can be made, we check if the amount of licenses already booked, the licenses
>  in use, the licenses reserved and the amount requested, are smaller or equal the license total.

Does it represent well the inequality?

backend/lm_backend/api/cruds/generic.py
julianaklulo: You're right, it's not needed so I removed it.

backend/lm_backend/api/cruds/generic.py
julianaklulo: My bad, I fixed it.

backend/lm_backend/api/routes/bookings.py
julianaklulo: We get request logging with FastAPI, not sure if we need more logging than what's already provided.

backend/lm_backend/api/routes/products.py
julianaklulo: That's a great idea, I updated the GenericCRUD to return the model name for each message.
Also updated the delete method to return the confirmation message.

backend/lm_backend/api/schemas.py
julianaklulo: Done!

backend/tests/api/routes/conftest.py
julianaklulo: Leftovers 🤦‍♀️

backend/tests/api/routes/conftest.py
julianaklulo: Leftovers 🤦‍♀️

backend/lm_backend/api/models.py
julianaklulo: Good catch, there were some more missing. I fixed them all.

backend/lm_backend/api/routes/clusters.py
julianaklulo: Fixed.

backend/lm_backend/api/routes/jobs.py
julianaklulo: Fixed.

backend/lm_backend/api/routes/jobs.py
julianaklulo: Fixed.

backend/tests/api/routes/test_clusters.py
julianaklulo: The name of the tests are self explanatory, I don't think we need doctrings for them. I removed these.

backend/lm_backend/api/routes/jobs.py
julianaklulo: We also need the exception in case the job is not in the list. Raising in the end makes it cover both cases (empty list and job not in the list).

backend/lm_backend/api/schemas.py
julianaklulo: Totally, I changed it to use PositiveInt.

backend/lm_backend/database.py
julianaklulo: I was not sure how to get the dialect once I removed the databases from this file. Do we really need it? From my tests it rendered the queries correctly.

backend/tests/conftest.py
julianaklulo: The test_app is used to instantiate the backend_client, which is then used by the tests.
We need the backend_client to inject the auth token, so invoking the app in each test wouldn't be feasible.

backend/lm_backend/api/models.py
julianaklulo: I broke this file in modules inside a `models/ `folder.

backend/lm_backend/api/schemas.py
julianaklulo: I broke it down into separate files.

backend/lm_backend/database.py
julianaklulo: We can do this change later, as it envolves changing the name in many places.

backend/tests/api/routes/test_bookings.py
julianaklulo: I added test cases for:
```
get by id, update, delete with bad parameter
update with bad data
```

backend/tests/conftest.py
dusktreader: We may revisit this in the future, but resolving for now.

dusktreader (approved):
Fantastic work! Way to stick with it through the long haul.

PR #227, opened 2023-06-16
[ASP-2211] Update agent to use new API

julianaklulo:
#### What
Update the agent to use the new API.

#### Why
To make it compatible with the new API.

`Task`: https://jira.scania.com/browse/ASP-2211

---

#### Peer Review


agent/lm_agent/backend_utils/utils.py
```diff
@@ -0,0 +1,357 @@
...
+
+async def check_backend_health():
+    """
+    Hit the API's health-check endpoint to make sure the API is available.
+    """
+    async with AsyncBackendClient() as backend_client:
+        resp = await backend_client.get("/lm/health")
+    if resp.status_code != 204:
+        logger.error("license-manager-backend health-check failed.")
+        raise LicenseManagerBackendConnectionError("Could not connect to the backend health-check endpoint")
```
matheushent: so, because our deployments are canary deployments, this would actually never happen. However, to make it realistic for people that use other type of deployments, I recommend setting up some retries here, as well as some timeout. This function can raise unrealistic behaviours if these parameters I mentioned cannot be set along side the load balancing parameters

agent/lm_agent/backend_utils/utils.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+async def get_jobs_from_backend() -> Optional[List[JobSchema]]:
+    """
+    Get all jobs for the cluster with its bookings from the backend.
+    """
+    cluster_data = await get_cluster_from_backend()
+    if cluster_data:
+        return cluster_data.jobs
+    return []
```
matheushent:
```suggestion
    return cluster_data.jobs if cluster_data else []
```

matheushent: actually, why the `if` is needed if there's only one response type defined in the `get_cluster_from_backend` annotation?

agent/lm_agent/backend_utils/utils.py
```diff
@@ -0,0 +1,357 @@
...
+
+def get_grace_times(cluster_data: ClusterSchema) -> Dict[int, int]:
+    """
+    Get the grace time for each feature_id in the cluster.
+    """
+    grace_times = {
+        feature.id: configuration.grace_time
+        for configuration in cluster_data.configurations
+        for feature in configuration.features
+    }
```
matheushent, line 220: Despite I think this code won't bring any performance problems **exclusively** due to the potential size of the database, I have a suggestion that I believe can improve the code base. Here it is the snippet I used to test two different approaches:

```python
import timeit
import itertools

pizzas = [{'pizza': 'margherita'}, {'pizza': 'hawaiian'}, {'pizza': 'pepperoni'}, {'pizza': 'veggie'}, {'pizza': 'meat lovers'}, {'pizza': 'vegan'}, {'pizza': 'supreme'}]
cheeses = [{'cheese': 'mozzarella'}, {'cheese': 'cheddar'}, {'cheese': 'parmesan'}]

def build_variants_with_compreehension():
    variants = {
        pizza.get("pizza"): cheese.get("cheese")
        for cheese in cheeses
        for pizza in pizzas
    }
    return variants

def build_variants_with_map():
    variants = dict(map(lambda x: (x[0]['pizza'], x[1]['cheese']), itertools.product(pizzas, cheeses)))
    return variants

time_to_run_comprehension = timeit.timeit(build_variants_with_compreehension, number=100000)
time_to_run_map = timeit.timeit(build_variants_with_map, number=100000)

print(f"Using map is {time_to_run_map/time_to_run_comprehension} times faster than comprehension")
```

agent/lm_agent/license_report.py
```diff
@@ -17,16 +18,18 @@
...
     for entry in license_configurations:
-        for feature in entry.features.keys():
-            if f"{entry.product}.{feature}" in local_licenses:
+        for feature in entry.features:
+            feature_name = feature.name
+            product_name = feature.product.name
+            if f"{product_name}.{feature_name}" in local_licenses:
                 filtered_entries.append(entry)
                 break
     return filtered_entries
```
matheushent, line 35: correct me if I understood wrong: this function returns a list with one element, right?

agent/lm_agent/reconciliation.py
```diff
@@ -38,91 +32,81 @@
...
-RECONCILE_URL_PATH = "/lm/api/v1/license/reconcile"
 
-
-def get_greatest_grace_time(job_id: str, grace_times: Dict[int, int], booking_rows: List) -> int:
+def get_greatest_grace_time_for_job(grace_times: Dict[int, int], job_bookings: List[BookingSchema]) -> int:
     """
-    Find the greatest grace_time for the given job_id.
+    Find the greatest grace_time for each feature booked by the given job_id.
     """
     greatest_grace_time = -1
```
matheushent, line 40: I feel like `float("-inf")` is safer here.

agent/lm_agent/reconciliation.py
```diff
@@ -38,91 +32,81 @@
...
 )
 
-RECONCILE_URL_PATH = "/lm/api/v1/license/reconcile"
 
-
-def get_greatest_grace_time(job_id: str, grace_times: Dict[int, int], booking_rows: List) -> int:
+def get_greatest_grace_time_for_job(grace_times: Dict[int, int], job_bookings: List[BookingSchema]) -> int:
     """
-    Find the greatest grace_time for the given job_id.
+    Find the greatest grace_time for each feature booked by the given job_id.
```
matheushent: I think the word `each` here is used wrongly because the function is returning the greatest grace time value between all features, not the greatest for each feature

agent/lm_agent/reconciliation.py
```diff
@@ -38,91 +32,81 @@
...
-            continue
-        for inner_book in book:
-            if str(inner_book["job_id"]) != str(job_id):
-                continue
-            config_id_for_grace_times = inner_book["config_id"]
-            greatest_grace_time = max(greatest_grace_time, grace_times[config_id_for_grace_times])
+    for booking in job_bookings:
+        feature_id = booking.feature_id
+        greatest_grace_time = max(greatest_grace_time, grace_times[feature_id])
     return greatest_grace_time
```
matheushent:
```suggestion
    return max(map(lambda feature_id: grace_times[feature_id], map(lambda booking: booking.feature_id, job_bookings)))
```

agent/lm_agent/backend_utils/models.py
```diff
@@ -0,0 +1,133 @@
+from enum import Enum
+from typing import List
+
+from pydantic import BaseModel, Field, PositiveInt
+
+from lm_agent.config import PRODUCT_FEATURE_RX
+
+
+class LicenseServerType(str, Enum):
```
dusktreader: This should probably go in a constants module instead of the models module since it isn't a model.

agent/lm_agent/reconciliation.py
matheushent: maybe generating a set with all the grace_time keys can improve as well

agent/lm_agent/backend_utils/utils.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+async def check_backend_health():
+    """
+    Hit the API's health-check endpoint to make sure the API is available.
+    """
+    async with AsyncBackendClient() as backend_client:
+        resp = await backend_client.get("/lm/health")
+    if resp.status_code != 204:
+        logger.error("license-manager-backend health-check failed.")
```
dusktreader: It would be good to log the actual response here. That way, we can see in the logs if it's a 400 or 500 or whatever. More breadcrumbs to follow.

agent/lm_agent/backend_utils/utils.py
```diff
@@ -0,0 +1,357 @@
...
+            f"/lm/features/{feature_id}/update_inventory",
+            json={
+                "total": total,
+                "used": used,
+            },
+        )
+        LicenseManagerBackendConnectionError.require_condition(
+            inventory_response.status_code == 200, f"Failed to update inventory: {inventory_response.text}"
+        )
+    return True
```
dusktreader: since this raises an exception when there is an error, do you need to have it return a boolean?

agent/lm_agent/backend_utils/utils.py
```diff
@@ -0,0 +1,357 @@
...
+        resp = await backend_client.delete(
+            f"lm/jobs/slurm_job_id/{slurm_job_id}/cluster_id/{cluster_data.id}"
+        )
+
+        LicenseManagerBackendConnectionError.require_condition(
+            resp.status_code == 200, f"Failed to remove job: {resp.text}"
+        )
+
+    logger.debug("##### Job removed successfully #####")
+    return True
```
dusktreader: Same as above. If an exception is being raised on failure, you don't need to return the boolean to indicate success of the method.

agent/lm_agent/backend_utils/utils.py
```diff
@@ -0,0 +1,357 @@
...
+            parsed_resp: List = job_response.json()["bookings"]
+
+    bookings = []
+
+    for booking in parsed_resp:
+        try:
+            parsed_booking = BookingSchema.parse_obj(booking)
+        except ValidationError as err:
+            logger.error(f"Could not validate booking data: {str(err)}")
+            raise LicenseManagerParseError("Could not parse booking data returned from the backend")
```
dusktreader: Since you are using this pattern a lot, you might look at using a `log_error()` helper method that you define once somewhere in the code base and use it wherever you want to use `handle_errors()`. Here's an example:


```python
from textwrap import dedent
from buzz import DoExceptParams

def log_error(params: DoExceptParams):
    message_template = dedent(
        """
        {final_message}
        Error:
        ______
        {err}
        Traceback:
        ----------
        {trace}
        """
    ).strip()
    message = message_template.format(
        final_message=params.final_message,
        err=str(params.err),
        trace="\n".join(format_tb(params.trace)),
    )
    logger.error(message)


for booking in parsed_resp:
    with LicenseManagerParseError.handle_errors("Could not parse booking data returned from the backend", do_except=log_error):
        bookings.append(parsed_booking)
```

In this example the traceback of the error is also included in the logs (which is very helpful for debugging). However, you could omit that and make the message template smaller.

If you only want the context manager to handle a specifc exception (or group of them), you can use the `handle_exc_class` parameter to `handle_errors`.

More information here: https://py-buzz.readthedocs.io/en/latest/features.html#exception-handling-context-manager

agent/lm_agent/reconciliation.py
```diff
@@ -38,91 +32,81 @@
...
-        if not book:
-            continue
-        for inner_book in book:
-            if str(inner_book["job_id"]) != str(job_id):
-                continue
-            config_id_for_grace_times = inner_book["config_id"]
-            greatest_grace_time = max(greatest_grace_time, grace_times[config_id_for_grace_times])
+    for booking in job_bookings:
+        feature_id = booking.feature_id
+        greatest_grace_time = max(greatest_grace_time, grace_times[feature_id])
```
dusktreader: You could use the iterable version of the max function and do this in one line:
```python
greatest_grace_time  = max(job_bookings, default=-1, key=lambda jb: grace_times[jb.feature_id])
```

agent/lm_agent/reconciliation.py
```diff
@@ -38,91 +32,81 @@
...
 
 def get_running_jobs(squeue_result: List) -> List:
     return [j for j in squeue_result if j["state"] == "RUNNING"]
 
 
-async def clean_booked_grace_time():
+async def clean_jobs_by_grace_time(cluster_data: ClusterSchema):
     """
-    Clean the booked licenses if the job's running time is greater than the grace_time.
+    Clean the jobs which running time is greater than the grace_time.
```
dusktreader:
```suggestion
    Clean the jobs where running time is greater than the grace_time.
```

agent/lm_agent/reconciliation.py
```diff
@@ -38,91 +32,81 @@
...
     squeue_running_jobs = get_running_jobs(squeue_result)
 
-    grace_times = await get_all_grace_times()
-    get_booked_call = []
-    for job in squeue_running_jobs:
-        job_id = job["job_id"]
-        get_booked_call.append(get_booking_for_job_id(job_id))
+    get_bookings_call = [get_bookings_for_job_id(job["job_id"]) for job in squeue_running_jobs]
+    results = await asyncio.gather(*get_bookings_call)
+    bookings_for_running_jobs = {job["job_id"]: result for job, result in zip(squeue_running_jobs, results)}
```
dusktreader, line 70: nice!

backend/lm_backend/api/routes/features.py
```diff
@@ -87,6 +87,31 @@ async def update_feature(
...
+        db_session=db_session, filter_field=Inventory.feature_id, filter_term=feature_id
+    )
+    if inventory:
+        await crud_inventory.update(
+            db_session=db_session,
+            id=inventory.id,
+            obj=inventory_update,
+        )
+
+    return await crud_feature.read(db_session=db_session, id=feature_id)
```
dusktreader, line 112: This is nicely done. Nice and succinct.

backend/lm_backend/api/routes/jobs.py
```diff
@@ -23,11 +27,28 @@
...
+            obj = {
+                "job_id": job_created.id,
+                "feature_id": booking.feature_id,
+                "quantity": booking.quantity,
+            }
+            try:
+                await crud_booking.create(db_session=db_session, obj=BookingCreateSchema(**obj))
+            except HTTPException:
+                await crud_job.delete(db_session=db_session, id=job_created.id)
+                raise
```
dusktreader: I think you could wrap the whole block (starting on line 39 with the try/except. Since you are going to blow it up if one fails, you don't need to handle the exception inside the loop.

dusktreader (changes requested):
Really nice work on this update! I left some comments for you, but I think this is really close.

agent/lm_agent/backend_utils/utils.py
julianaklulo: What kind of unrealistic behaviours can happen here? This function has been there for a while.
If the backend doesn't respond, we want to abort the reconciliation, since it doesn't make sense to try updating the backend if the agent can't connect to it.

agent/lm_agent/backend_utils/utils.py
julianaklulo: Adding the `if` solved some mypy issue that was happening, but it was related to the schema (the `jobs` field was `Optional`, which could return `None`, but it will always be an empty list if there are no jobs).
Changing the schema fixed it, so I tried it now and it works without that `if` statement, thanks!

agent/lm_agent/license_report.py
julianaklulo: Each configuration can have `n` features attached to it.
This function checks if any of the features in the configuration are in the cluster. If one of them is in the cluster, we need the entire configuration.
So it should create a list of configurations, not necessarily one element.

agent/lm_agent/reconciliation.py
julianaklulo: Fixed it, thanks!

agent/lm_agent/reconciliation.py
julianaklulo: I feel like using `map` here would reduce the readability.
The unit test for this function failed with `max() arg is an empty sequence` using the suggested change.
Looks like it doesn't cover the case where the job doesn't have any bookings.

agent/lm_agent/backend_utils/models.py
julianaklulo: Done!

agent/lm_agent/backend_utils/utils.py
julianaklulo: Done!

agent/lm_agent/backend_utils/utils.py
julianaklulo: Nope, my bad. I removed it.

agent/lm_agent/backend_utils/utils.py
julianaklulo: Removed.

agent/lm_agent/reconciliation.py
julianaklulo: Fixed!

backend/lm_backend/api/routes/jobs.py
julianaklulo: That makes sense, I updated it.

agent/lm_agent/backend_utils/utils.py
julianaklulo: Great suggestion, thanks! I implemented it in the `logs` module and used it here.

agent/lm_agent/reconciliation.py
julianaklulo: This suggestion makes `greatest_grace_time` receive a `job_booking` instance (or `-1` by default if no `job_bookings` are passed).
I updated it to make the validation to return the `grace_time` and it worked, thanks!

agent/lm_agent/reconciliation.py
julianaklulo: I went with @dusktreader's [suggestion](https://github.com/omnivector-solutions/license-manager/pull/227#discussion_r1237670454) for this part of the code.

agent/lm_agent/backend_utils/utils.py
julianaklulo: I feel that the improvement in the execution time is not worth decreasing the readability.

matheushent: don't you think that `dict(map(lambda x: (x[0]['pizza'], x[1]['cheese']), itertools.product(pizzas, cheeses)))` is readable at all?

agent/lm_agent/backend_utils/utils.py
matheushent: you can disregard my comment since the cases I'm thinking are not relevant now

agent/lm_agent/backend_utils/utils.py
dusktreader: Using comprehensions is generally best practice in python. We should prefer this approach unless the performance gain is very substatial.

dusktreader (approved):

PR #234, opened 2023-07-17
PENG-1510 Fixed feature payload to include booked total

dusktreader:
#### What
Added a computed "booked_total" column to the feature result payload that sums up the bookings quantities for a feature.

#### Why
To better support the UI and CLI so that secondary requests do not need to be made to get all the bookings and then correlate those results with the features.

`Task`: https://app.clickup.com/t/18022949/PENG-1510

---

#### Peer Review

### DIFF
--- a/backend/lm_backend/api/models/feature.py
+++ b/backend/lm_backend/api/models/feature.py
@@ -1,8 +1,10 @@
 """Database model for Features."""
-from sqlalchemy import Column, Integer, String
+from sqlalchemy import Column, Integer, SQLColumnExpression, String, func, select
+from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import relationship
 from sqlalchemy.sql.schema import CheckConstraint, ForeignKey
 
+from lm_backend.api.models.booking import Booking
 from lm_backend.database import Base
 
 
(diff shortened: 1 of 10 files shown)


backend/lm_backend/api/models/feature.py
```diff
@@ -30,11 +32,27 @@ class Feature(Base):
...
+        """
+        return sum(b.quantity for b in self.bookings)
+
+    @booked_total.inplace.expression
+    @classmethod
+    def _booked_total_expression(cls) -> SQLColumnExpression[int]:
+        """
+        Provide a sql expression for computing the total bookings for each feature in a subquery.
+        """
+        return select(func.sum(Booking.quantity)).where(Booking.feature_id == cls.id).label("booked_total")
```
julianaklulo, line 48: Where is this query executed?

julianaklulo (commented):
The logic here is very clever, nice!
~I tested it locally against my test database but `booked_total` is zero.~
~I'm wondering if the query is being run, since I couldn't locate it in the code.~
(the agent had deleted the booking 🤦‍♀️)

backend/lm_backend/api/models/feature.py
dusktreader: This isn't being used currently, however, if you ever needed to use the booked_total inside of a query, then this would provide the sql logic automatically. Without it, the query expression wouldn't know how to use booked_total.

With hybrid attributes, it's common to supply an expression for instances where the hybrid attribute needs to be used in a query.

This was leftovers from when I was working on getting this working without using a lazy loader. However, the logic is sound and tested. I think it would be good to leave this in.

julianaklulo (approved):

PR #246, opened 2023-08-04
Remove clusters

julianaklulo:
#### What
Remove cluster entity from the entire codebase.

#### Why
To reduce complexity and facilitate UI integration.

---

#### Peer Review


agent/lm_agent/backend_utils/utils.py
```diff
@@ -320,24 +275,49 @@ async def get_bookings_for_job_id(slurm_job_id: str) -> List[BookingSchema]:
...
+            parsed_resp: List = feature_response.json()
+
+    features = []
+
+    for feature in parsed_resp:
+        with LicenseManagerParseError.handle_errors(
+            "Could not parse feature data returned from the backend", do_except=log_error
+        ):
+            parsed_feature = FeatureSchema.parse_obj(feature)
+        features.append(parsed_feature)
```
dusktreader: Since you are aborting on the first parsing error encountered, you could put the `handle_errors` context manager a level higher and do the list construction with a comprehension:

```python
with LMPError.handle_errors(...):
    features = [FeatureSchema.parse_obj(f) for f in parsed_resp]
```

agent/lm_agent/backend_utils/utils.py
```diff
@@ -320,24 +275,49 @@ async def get_bookings_for_job_id(slurm_job_id: str) -> List[BookingSchema]:
...
+    booking_sum = 0
 
-        if feature_id is not None:
-            for job in cluster.jobs:
-                for booking in job.bookings:
-                    if booking.feature_id == feature_id:
-                        booking_sum[cluster.id] = booking_sum.get(cluster.id, 0) + booking.quantity
+    for feature in features:
+        if f"{feature.product.name}.{feature.name}" == product_feature:
+            booking_sum += feature.booked_total
```
dusktreader: You could use the `sum()` function here with a comprehension:
```python
booking_sum = sum(f.booked_total for f in features if f"{f.product.name}.{f.name}" == product_feature)
```

backend/alembic/versions/0001--implement_new_data_struture_schema.py
dusktreader, line 1: We should combine the migration scripts into one final migration for the data restructure.

backend/lm_backend/api/routes/jobs.py
```diff
@@ -49,7 +70,29 @@ async def create_job(
...
 @router.get(
-    "/",
+    "/by_client_id",
+    response_model=List[JobSchema],
+    status_code=status.HTTP_200_OK,
+)
+async def read_jobs_by_client_id(
+    secure_session: SecureSession = Depends(secure_session(Permissions.JOB_VIEW)),
+):
+    """Return the jobs with the specified client_id."""
```
dusktreader: Since the `client_id` isn't specified in the request itself, your docstring should specify that it uses the client id from the identity payload in the request.

backend/lm_backend/api/routes/jobs.py
```diff
@@ -96,55 +139,69 @@ async def delete_job(
...
-        db_session=secure_session.session, search=slurm_job_id, force_refresh=True
+    client_id = secure_session.identity_payload.client_id
+
+    if not client_id:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=("Couldn't find a valid client_id in the access token."),
+        )
+
+    jobs: List[Job] = await crud_job.filter(
```
dusktreader, line 163: could you add a filter term for `cluster_client_id` to this filter method?

backend/lm_backend/api/routes/utils.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    features: List[Feature] = await crud_feature.filter(
+        db_session=db_session, filter_field=Feature.name, filter_term=feature_name
+    )
+
+    for feature in features:
+        configuration = await crud_configuration.read(db_session=db_session, id=feature.config_id)
+
+        if configuration:
+            if configuration.cluster_client_id == client_id:
```
dusktreader: I think it would be better to just make this a single `if` with an `and`

backend/lm_backend/api/schemas/feature.py
```diff
@@ -31,6 +31,16 @@ class FeatureUpdateSchema(BaseUpdateSchema):
     used: Optional[int] = None
 
 
+class FeatureUpdateByNameSchema(BaseUpdateSchema):
+    """
+    Represents the features in a feature configuration.
```
dusktreader: I think this docstring isn't right.

dusktreader (changes requested):
Nice work! Left a few comments for you to consider.

backend/lm_backend/api/schemas/feature.py
julianaklulo: Is `"Represents the feature usage data that will be updated using the name as a filter."` better?

backend/lm_backend/api/routes/jobs.py
julianaklulo: I updated the CRUD filter method to accept a list of expressions to filter.
Now it's possible filter using both parameters at the same time.

backend/lm_backend/api/schemas/feature.py
dusktreader: Yes, thank you.

dusktreader (approved):

PR #249, opened 2023-08-10
[PENG-1588] Update configuration create route

julianaklulo:
#### What
Update `POST /configurations` route to accept a configuration with features and license servers included.

#### Why
To make it easier to create new configurations.

`Task`: https://app.clickup.com/t/18022949/PENG-1588

---

#### Peer Review

### DIFF
--- a/backend/lm_backend/api/routes/configurations.py
+++ b/backend/lm_backend/api/routes/configurations.py
@@ -4,18 +4,28 @@ from fastapi import APIRouter, Depends, HTTPException, Query, status
 
 from lm_backend.api.cruds.generic import GenericCRUD
 from lm_backend.api.models.configuration import Configuration
+from lm_backend.api.models.feature import Feature
+from lm_backend.api.models.license_server import LicenseServer
+from lm_backend.api.models.product import Product
 from lm_backend.api.schemas.configuration import (
+    ConfigurationCompleteCreateSchema,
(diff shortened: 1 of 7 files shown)


backend/lm_backend/api/routes/configurations.py
```diff
@@ -24,11 +34,54 @@
...
+                    "name": feature.name,
+                    "product_id": feature.product_id,
+                    "config_id": configuration_created.id,
+                    "reserved": feature.reserved,
+                }
+                await crud_feature.create(
+                    db_session=secure_session.session, obj=FeatureCreateSchema(**feature_obj)
+                )
+        except HTTPException:
+            await crud_configuration.delete(db_session=secure_session.session, id=configuration_created.id)
```
dusktreader, line 64: I don't think you need to explicitly delete. If an exception is raised, the whole transaction should be rolled back including any rows created.

dusktreader (changes requested):
Looks good. I just think we don't need the explicit delete on create failure.

backend/lm_backend/api/routes/configurations.py
julianaklulo: I removed the delete but the configuration row is still there after the failure case.
It looks like the creation is being done in different transactions.

dusktreader (approved):

PR #257, opened 2023-08-23
Add feature bulk update

julianaklulo:
#### What
Add an endpoint to bulk update features.

#### Why
The agent was making `n` requests to update `n` features, with the new endpoint it's reduced to only one request.

`Task`: https://app.clickup.com/t/18022949/PENG-1613

---

#### Peer Review

### DIFF
--- a/backend/lm_backend/api/routes/features.py
+++ b/backend/lm_backend/api/routes/features.py
@@ -108,6 +108,49 @@ async def update_feature_by_client_id(
 
 
 @router.put(
+    "/bulk",
+    status_code=status.HTTP_200_OK,
+)
+async def bulk_update_feature(
+    features: List[FeatureUpdateByNameSchema],
+    secure_session: SecureSession = Depends(secure_session(Permissions.FEATURE_EDIT)),
+):
+    """
+    Update a list of features in the database using the name of each feature.
+    Since the name is not unique across clusters, the client_id
+    in the token is used to identify the cluster.
+    """
+    client_id = secure_session.identity_payload.client_id
+
+    if not client_id:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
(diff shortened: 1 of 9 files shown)


backend/lm_backend/api/routes/features.py
```diff
@@ -107,6 +107,49 @@ async def update_feature_by_client_id(
...
+    client_id = secure_session.identity_payload.client_id
+
+    if not client_id:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=("Couldn't find a valid client_id in the access token."),
+        )
+
+    for feature_update in features:
+        feature_id = await find_feature_id_by_name_and_client_id(
```
dusktreader: This is still going to be inefficient because you're going to be doing a `select` query for each feature individually. To speed it up, you could make a method that gets all the features to update using a query like:

```sql
select *
from features
join configuration
where (feature.name, configuration.cluster_client_id) in values
   ('foo', 1),
   ('bar', 2)
```

in sqlalchemy this would look something like
```python
query = (
    select(Feature)
    .join(Configuration)
    .where(
        tuple_(
            Feature.name, 
            Configuration.cluster_client_id
        ).in_(
            [
                ("foo", 1),
                ("bar", 2),
            ]
        )
    )
)
```

dusktreader (changes requested):
Looks good, but I have a suggestion to speed it up even more.

backend/lm_backend/api/routes/features.py
julianaklulo: Thank you for the suggestion. I ended up creating a `Feature CRUD` for and added the `bulk_update` method there, using the query you wrote.
The approach is similar with what we did with the `Booking CRUD`, to have a query for creating a booking.

dusktreader (approved):

PR #268, opened 2023-09-13
[ASP-3746] Improve update route

julianaklulo:
#### What
Improve the update route to:

* Create feature if it's in the payload and doesn't have an id
* Update feature if it's in the payload and has an id
* Delete any feature not present in the payload

The same applies to license servers.

#### Why
To make it easier to update all resources related to the configuration in a single request.

`Task`: https://jira.scania.com/browse/ASP-3746

---

#### Peer Review

### DIFF
--- a/backend/lm_backend/api/cruds/configuration.py
+++ b/backend/lm_backend/api/cruds/configuration.py
@@ -1,0 +1,56 @@
+"""Feature CRUD class for SQLAlchemy models."""
+from fastapi import HTTPException
+from loguru import logger
+from sqlalchemy import delete
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from lm_backend.api.cruds.generic import GenericCRUD
+from lm_backend.api.models.feature import Feature
+from lm_backend.api.models.license_server import LicenseServer
+from lm_backend.api.schemas.configuration import ConfigurationCompleteUpdateSchema
+
+
+class ConfigurationCRUD(GenericCRUD):
+    """Configuration CRUD module to implement configuration update."""
+
+    async def delete_features(
+        self, db_session: AsyncSession, configuration_id: int, payload: ConfigurationCompleteUpdateSchema
+    ):
+        """
+        Delete all features that aren't in the payload.
+        """
+        features_in_payload = [feature.id for feature in payload.features] if payload.features else []
+        delete_unused_features_query = (
+            delete(Feature)
+            .where(~Feature.id.in_(features_in_payload))
+            .where(Feature.config_id == configuration_id)
+        )
+
+        try:
+            await db_session.execute(delete_unused_features_query)
+        except Exception as e:
+            logger.error(e)
+            raise HTTPException(status_code=400, detail="Feature could not be deleted.")
+
+    async def delete_license_servers(
+        self, db_session: AsyncSession, configuration_id: int, payload: ConfigurationCompleteUpdateSchema
+    ):
+        """
+        Delete all license servers that aren't in the payload.
+        """
+        license_servers_in_payload = (
+            [license_server.id for license_server in payload.license_servers]
+            if payload.license_servers
+            else []
+        )
+        delete_unused_license_servers_query = (
+            delete(LicenseServer)
+            .where(~LicenseServer.id.in_(license_servers_in_payload))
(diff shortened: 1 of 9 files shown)


backend/lm_backend/api/cruds/configuration.py
```diff
@@ -0,0 +1,56 @@
...
+        self, db_session: AsyncSession, configuration_id: int, payload: ConfigurationCompleteUpdateSchema
+    ):
+        """
+        Delete all license servers that aren't in the payload.
+        """
+        license_servers_in_payload = (
+            [license_server.id for license_server in payload.license_servers]
+            if payload.license_servers
+            else []
+        )
```
dusktreader: Could shorten this a little more even:

```suggestion
        license_servers_in_payload = [ls.id for ls in payload.license_servers or []]
```

backend/lm_backend/api/cruds/configuration.py
```diff
@@ -0,0 +1,56 @@
...
+        delete_unused_license_servers_query = (
+            delete(LicenseServer)
+            .where(~LicenseServer.id.in_(license_servers_in_payload))
```
dusktreader, line 46: Or, instead of above suggestion, just apply the logic here:

```suggestion
            .where(~LicenseServer.id.in_(payload.license_servers or []))
```

backend/lm_backend/api/routes/configurations.py
```diff
@@ -149,15 +151,93 @@ async def read_configuration(
...
-    return await crud_configuration.update(
-        db_session=secure_session.session,
-        id=configuration_id,
-        obj=configuration_update,
-    )
+    """
+    Update a configuration in the database.
+
+    If there are features in the payload, they'll be updated if they have an id,
+    or will be created if the id is None.
```
dusktreader: For a compound sentence, you need to have a subject in the second independent clause. Alternatively, you could make the sentence have a compound predicate instead by removing the comma:

Compound sentence:
```suggestion
    If there are features in the payload, they'll be updated if they have an id,
    or they will be created if the id is None.
```

Compound predicate:
```suggestion
    If there are features in the payload, they'll be updated if they have an id
    or will be created if the id is None.
```

backend/lm_backend/api/routes/configurations.py
```diff
@@ -149,15 +151,93 @@ async def read_configuration(
...
+                        "name": feature_update.name,
+                        "product_id": feature_update.product_id,
+                        "config_id": configuration_id,
+                        "reserved": feature_update.reserved,
+                    }
+                    await crud_feature.create(
+                        db_session=secure_session.session, obj=FeatureCreateSchema(**feature_obj)
+                    )
+        except HTTPException:
+            raise
```
dusktreader: This is effectively a no-op. Shoudn't need the try/except here at all.

backend/lm_backend/api/routes/configurations.py
```diff
@@ -149,15 +151,93 @@ async def read_configuration(
...
+                    license_server_obj = {
+                        "config_id": configuration_id,
+                        "host": license_server_update.host,
+                        "port": license_server_update.port,
+                    }
+                    await crud_license_server.create(
+                        db_session=secure_session.session, obj=LicenseServerCreateSchema(**license_server_obj)
+                    )
+        except HTTPException:
+            raise
```
dusktreader: also an effective no-op.

dusktreader (changes requested):
A few style and grammar (for text) suggestions for you.

backend/lm_backend/api/cruds/configuration.py
julianaklulo: I went with the suggestions above.

backend/lm_backend/api/routes/configurations.py
julianaklulo: Thanks, I updated it to a compound sentence.

backend/lm_backend/api/cruds/configuration.py
julianaklulo: Mypy didn't like it, the explict `"if payload.license_servers"` ensures that it's an iterable and not None.

dusktreader (approved):

PR #276, opened 2023-09-27
[ASP-3906] Add constraints to schemas

julianaklulo:
#### What
Update the API schemas to include constraints in all numerical values, also add max lenght to string values.

#### Why
To ensure the configurations will be created with reasonable values.

`Task`: https://jira.scania.com/browse/ASP-3906

---

#### Peer Review

### DIFF
--- a/backend/lm_backend/api/schemas/booking.py
+++ b/backend/lm_backend/api/schemas/booking.py
@@ -1,7 +1,7 @@
 """Booking schemas for the License Manager API."""
 from typing import Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field, PositiveInt
 
 from lm_backend.api.schemas.base import BaseCreateSchema, BaseUpdateSchema
 
@@ -11,9 +11,11 @@ class BookingCreateSchema(BaseCreateSchema):
     Represents the booking of a feature.
     """
 
-    job_id: int
-    feature_id: int
-    quantity: int
+    job_id: int = Field(..., title="Job ID", description="The ID of the job that booked the feature.")
+    feature_id: int = Field(..., title="Feature ID", description="The ID of the feature that was booked.")
+    quantity: PositiveInt = Field(
+        ..., title="Quantity", description="The quantity of the feature that was booked."
+    )
 
 
 class BookingUpdateSchema(BaseUpdateSchema):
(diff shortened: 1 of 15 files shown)


backend/lm_backend/api/schemas/booking.py
```diff
@@ -11,30 +11,40 @@ class BookingCreateSchema(BaseCreateSchema):
     Represents the booking of a feature.
     """
 
-    job_id: int
-    feature_id: int
-    quantity: int
+    job_id: int = Field(..., title="Job ID", description="The ID of the job that booked the feature.")
+    feature_id: int = Field(..., title="Feature ID", description="The ID of the feature that was booked.")
+    quantity: PositiveInt = Field(
```
dusktreader, line 16: Should there be an upper bound on this one?

backend/lm_backend/api/schemas/license_server.py
```diff
@@ -23,39 +27,47 @@ class LicenseServerWithoutConfigIdCreateSchema(BaseCreateSchema):
     The config_id will be handled in the endpoint.
     """
 
-    host: str
-    port: PositiveInt
+    host: str = Field(..., title="Host", description="The host of the license server.")
+    port: PositiveInt = Field(..., title="Port", description="The port of the license server.")
```
dusktreader, line 45: This definitely should have an upper bound (65535).

backend/lm_backend/api/schemas/license_server.py
```diff
@@ -23,39 +27,47 @@ class LicenseServerWithoutConfigIdCreateSchema(BaseCreateSchema):
     The config_id will be handled in the endpoint.
     """
 
-    host: str
-    port: PositiveInt
+    host: str = Field(..., title="Host", description="The host of the license server.")
```
dusktreader, line 44: We should also make sure valid hostnames are supplied here. According to https://www.rfc-editor.org/rfc/rfc1123: [a-z0-9-.]+ (basically)

dusktreader (changes requested):
Some small nitpicky ones for you.

backend/lm_backend/api/schemas/license_server.py
julianaklulo: > "Host software MUST handle host names of up to 63 characters and SHOULD handle host names of up to 255 characters."

This regex would consider the 63 characters limit: `"^(?!-)[A-Za-z0-9-]{1,63}(?<!-)(\.[A-Za-z0-9-]{1,63})*$"`

backend/lm_backend/api/schemas/booking.py
julianaklulo: The quantity is validaded in the CRUD create booking method.
The method calculates the quantity of licenses free to decide if the booking can be created.
I don't think this is something that can be handled in the schema.

backend/lm_backend/api/schemas/license_server.py
julianaklulo: Added a validator for it.

dusktreader (approved):

PR #284, opened 2023-10-11
patch code to avoid committing against the database in GET endpoints

matheushent:
#### What
This PR applies a simple patch in the API code to avoid running `session.commit()` against the database while only read queries are run.

#### Why
Experimental fix for the performance issues currently on production

---

#### Peer Review


dusktreader (approved):

agent/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "license-manager-agent"
-version = "3.0.9"
+version = "3.0.10-alpha.1"
```
julianaklulo: Should this be `3.0.10a1`?

backend/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "license-manager-backend"
-version = "3.0.9"
+version = "3.0.10-alpha.1"
```
julianaklulo: Same here

lm-cli/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "license-manager-cli"
-version = "3.0.9"
+version = "3.0.10-alpha.1"
```
julianaklulo: And here

agent/pyproject.toml
matheushent: there's no standard across our services actually. Some use this nomenclature, other use `ax`. Anyway for me, I will change

PR #338, opened 2024-04-18
[ASP-4902] Delete bookings on usage matching

julianaklulo:
#### What
Refactor of the booking and job cleaning process.

#### Why
Previously the bookings were cleaned only when the grace time of their parent job had expired.

Now, the jobs and the bookings can be cleaned when:
* the job doesn't have any bookings
* the job is not running anymore
* the grace time for the job expired
* the license booked has been checked out from the license server

`Task`: https://jira.scania.com/browse/ASP-4902

---

#### Peer Review


lm-agent/lm_agent/server_interfaces/flexlm.py
```diff
@@ -50,22 +50,29 @@ async def get_output_from_server(self, product_feature: str):
...
+    async def get_report_item(self, feature_id: int, product_feature: str):
         """Override abstract method to parse FlexLM license server output into License Report Item."""
 
         server_output = await self.get_output_from_server(product_feature)
         parsed_output = self.parser(server_output)
 
         # raise exception if parser didn't output license information
         if any(
-            [parsed_output is None, parsed_output.get("total") is None, parsed_output.get("used") is None]
+            [
```
dusktreader: This might be a useful place to use `check_expressions` from the py-buzz package.
https://github.com/dusktreader/py-buzz/blob/main/buzz/tools.py#L130

dusktreader (approved):
Really nice work here! +1

PR #339, opened 2024-04-24
[ASP-5048] Fixes found during integration testing of ASP-4902

julianaklulo:
#### What
Fix issues found during integration testing of `[ASP-4902] Implement algorithm to delete bookings`:
* Fix inconsistent usage of `user_name` vs `username` in parsers and reports
* Split the hostname in the license usage report to match the Slurm instance id from the booking
* Improve logging in the new `clean_jobs_and_bookings` service
* Improve logging in the `prolog` and `epilog` scripts

#### Why
These issues were found during the integration testing in QA environment and were preventing the solution from working as expected.

`Task`: https://jira.scania.com/browse/ASP-5048

---

#### Peer Review


lm-agent/lm_agent/services/clean_jobs_and_bookings.py
```diff
@@ -67,7 +67,7 @@ def extract_usages_from_report(report_item: LicenseReportItem) -> List[Extracted
         ExtractedUsageSchema(
             feature_id=report_item.feature_id,
             username=usage["username"],
-            lead_host=usage["lead_host"],
+            lead_host=usage["lead_host"].split(".")[0],
```
dusktreader: Should probably add a comment explaining why this logic is being applied here.

lm-agent/lm_agent/services/clean_jobs_and_bookings.py
```diff
@@ -117,10 +117,12 @@ async def clean_jobs_without_bookings(cluster_jobs: List[JobSchema]) -> List[Job
     logger.debug("##### Cleaning jobs without bookings")
 
     jobs_to_delete_call = []
+    jobs_to_delete = []
 
     for job in cluster_jobs[:]:
         if not job.bookings:
             jobs_to_delete_call.append(remove_job_by_slurm_job_id(job.slurm_job_id))
+            jobs_to_delete.append(job.slurm_job_id)
```
dusktreader, line 127: Instead of having two lists here, why not accumulate the slurm_job_id values in a list and then use a comprehension to produce the list of async calls to make? Something like:

```
    jobs_to_delete[]
    for job in cluster_jobs[:]:
        if not job.bookings:
            jobs_to_delete.append(job.slurm_job_id)
            cluster_jobs.remove(job)
    ...     
    await asyncio.gather(*[remove_job_by_slurm_job_id(jid) for jid in jobs_to_delete])
    ```

**Review by dusktreader — CHANGES_REQUESTED** [review]

File: lm-agent/lm_agent/services/clean_jobs_and_bookings.py (hunk 1 above)
Comment (julianaklulo) [reply lm-agent/lm_agent/services/clean_jobs_and_bookings.py]: Done!

File: lm-agent/lm_agent/services/clean_jobs_and_bookings.py (hunk 2 above)
Comment (julianaklulo) [reply lm-agent/lm_agent/services/clean_jobs_and_bookings.py:127]: Good suggestion, thanks! I updated it.

File: lm-agent/tests/parsing/test_flexlm.py [hunk 3]
```diff
@@ -44,35 +46,43 @@ def test_parse_feature_line(line, result):
...
-                "lead_host": "myserver.example.com",
-                "booked": 29,
-            },
+            LicenseUsesItem(
+                **{
+                    "username": "user1",
+                    "lead_host": "myserver.example.com",
+                    "booked": 29,
+                }
+            ),
```
dusktreader: Why use a dict here instead of just using kwargs?

dusktreader (approved):
One small comment for you which you can take or leave. Otherwise, looks good!

lm-agent/tests/parsing/test_flexlm.py
julianaklulo: The tests were using the dicts before, so it was faster to update them this way.
But since that may difficult the readabily, I changed it now to kwargs.

PR #344, opened 2024-04-30
[ASP-5048]  Fix lead host parsing when the job runs in multiple nodes

julianaklulo:
#### What
Fix the extraction of the lead host when the job runs in multiple nodes.

#### Why
The nodelist was being splitted by `,`, but the nodelist comes inside square brackets, which was not returning the expected values.

Example: `node-[1,2,3]` was being parsed as `node-[1` instead of `node-1`

`Task`: https://jira.scania.com/browse/ASP-5048

---

#### Peer Review

### DIFF
--- a/lm-agent/lm_agent/workload_managers/slurm/cmd_utils.py
+++ b/lm-agent/lm_agent/workload_managers/slurm/cmd_utils.py
@@ -7,13 +7,13 @@ from typing import Dict, List, Optional, Union
 
 from lm_agent.backend_utils.models import LicenseBooking
 from lm_agent.logs import logger
-from lm_agent.workload_managers.slurm.common import (
-    CMD_TIMEOUT,
-    ENCODING,
-    SACCTMGR_PATH,
-    SCONTROL_PATH,
-    SQUEUE_PATH,
-)
+
+
+SCONTROL_PATH = "/usr/bin/scontrol"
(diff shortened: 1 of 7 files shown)


lm-agent/lm_agent/workload_managers/slurm/cmd_utils.py
```diff
@@ -145,6 +145,30 @@ async def scontrol_show_lic():
...
+    ]
+
+    proc = await asyncio.create_subprocess_shell(
+        shlex.join(cmd), stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.STDOUT
+    )
+
+    stdout, _ = await asyncio.wait_for(proc.communicate(), CMD_TIMEOUT)
+    output = str(stdout, encoding=ENCODING)
+
+    return output.split("\n")[0]
```
fschuch: Optional comment: It may be nice to raise some error if it results in an empty string

fschuch (approved):
Nice work! I approved it as it is, but left an optional suggestion.

lm-agent/lm_agent/workload_managers/slurm/cmd_utils.py
julianaklulo: Good catch, I used `require_condition` from `buzz` to raise an exception if the `lead_host` is empty and also added a test for this part.

PR #357, opened 2024-06-13
PENG-2281 upgrade pydantic settings to ^2.7.4 in the agent project

matheushent:
#### What
This PR upgrades the pydantic version in the agent project, as well as adds *pydantic-settings* as a dependency. Notice that some unit tests have been slightly changed: this was needed because the way comparison works between pydantic models and dictionaries in the most recent version.

#### Why
Keep the project up to date with the most recent technologies

`Task`: https://sharing.clickup.com/t/h/c/18022949/PENG-2281/S5C5TX46EBVEU8Z

---

#### Peer Review


matheushent:
I have no idea what's going on with the CI. Locally, the CI passes with no error
image

lm-agent/lm_agent/backend_utils/models.py
```diff
@@ -79,13 +79,15 @@ class JobSchema(BaseModel):
 
     bookings: List[BookingSchema] = []
 
+    model_config = ConfigDict(from_attributes=True)
```
dusktreader: Why does this need a `from_attributes=True`? Shouldn't this only be set if `orm_mode` was set before?

lm-agent/lm_agent/server_interfaces/license_server_interface.py
```diff
@@ -1,13 +1,12 @@
...
-import abc
 from typing import List
 
 from pydantic import BaseModel, Field
 
 from lm_agent.config import PRODUCT_FEATURE_RX
 
 
-class LicenseServerInterface(metaclass=abc.ABCMeta):
+class LicenseServerInterface:
```
dusktreader, line 9: Why remove the ABC markers from this class?

dusktreader (changes requested):

lm-agent/lm_agent/services/license_report.py
```diff
@@ -63,7 +63,7 @@ async def report() -> typing.List[LicenseReportItem]:
     logger.debug("### Licenses in the cluster: ")
     logger.debug(filtered_entries)
 
-    license_server_interface: LicenseServerInterface
+    # license_server_interface: LicenseServerInterface
```
julianaklulo: Why is this commented?

julianaklulo (changes requested):

lm-agent/lm_agent/backend_utils/models.py
matheushent: `orm_mode` doesn't exist in pydantic v2, instead, just setting `from_attributes` should work ([reference](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.from_attributes))

lm-agent/lm_agent/server_interfaces/license_server_interface.py
matheushent: because it doesn't need to and mypy complains about it

lm-agent/lm_agent/services/license_report.py
matheushent: oh, it's a left over. But essentially, mypy complains with the definition of this variable the way it is

dusktreader (approved):

lm-agent/lm_agent/backend_utils/models.py
julianaklulo: But this schema didn't have the `orm_mode` set before, since the agent doesn't interface with the database.
It doesn't need to have `from_attributes=True`, the same way the other schemas in the same module don't have it.

matheushent: well, I'm not sure how it worked before, but now I see this model is used for calling `model_validate`, which needs the `from_attributes=True`

julianaklulo: According to [Pydantic's documentation](https://docs.pydantic.dev/latest/concepts/models/#arbitrary-class-instances):

> "Pydantic models can also be created from arbitrary class instances by reading the instance attributes corresponding to the model field names. One common application of this functionality is integration with object-relational mappings (ORMs)."

Here we're not creating it from a class. Notice that all instances of `model_validate` is converting a response from the API (JSON converted to a `dict`) into a Pydantic model. All the other schemas (ConfigurationSchema, FeatureSchema and LicenseBookingRequest) are doing the same without `from_attributes=True` being set.

dusktreader: It is true that you do not need `from_attributes=True` to use `model_validate`.

matheushent: adjusted in [55c783f](https://github.com/omnivector-solutions/license-manager/pull/357/commits/55c783f65a6ffc5d310636f2eca6c6f8013e74fa)

julianaklulo (approved):

PR #368, opened 2024-07-24
[ASP-5025] Refactor of License Manager Simulator to fix FK issue

julianaklulo:
#### What
Refact the License Manager Simulator:
* Changed the `engine` and `session` to be Async
* Fixed `models` to add cascade delete on the `LicenseInUse` (solving the FK issue)
* Updated `CRUD` module to use the new `session`
* Updated `main` module to remove trailing slash and use the new `CRUD` methods
* Updated unit tests and improved test coverage
* Updated test job application to use the new routes

#### Why
When deleting a `License`, the `LicenseInUse` related to it was kept in the database. It was not possible to delete these entries because the parent `License` was not available anymore.

`Task`: https://jira.scania.com/browse/ASP-5025

---

#### Peer Review

### DIFF
--- a/lm-simulator/lm_simulator/crud.py
+++ b/lm-simulator/lm_simulator/crud.py
@@ -1,129 +1,155 @@
 from typing import List
 
-from sqlalchemy import delete, select
+from buzz import enforce_defined, handle_errors, require_condition
+from fastapi import HTTPException, status
+from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from lm_simulator.models import License, LicenseInUse
 from lm_simulator.schemas import LicenseCreate, LicenseInUseCreate, LicenseInUseRow, LicenseRow
 
 
-class LicenseNotFound(Exception):
-    """The requested license is not found."""
+async def add_license(session: Session, license: LicenseCreate) -> LicenseRow:
+    """
+    Add a new License to the database.
+    """
+    db_license = License(**license.model_dump())
+
+    with handle_errors(
+        "Can't create License, check the input data",
+        raise_exc_class=HTTPException,
+        exc_builder=lambda exc_class, msg: exc_class(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Can't create License, check the input data",
+        ),
+    ):
+        session.add(db_license)
+        await session.commit()
+        await session.refresh(db_license)
+
+        return LicenseRow.model_validate(db_license, from_attributes=True)
 
 
-class NotEnoughLicenses(Exception):
-    """The number of requested licenses is bigger than the available."""
+async def list_licenses(session: Session) -> List[LicenseRow]:
+    """
+    List all Licenses in the database.
+    """
+    query = await session.execute(select(License))
+    db_licenses = query.scalars().all()
+    return [LicenseRow.model_validate(license) for license in db_licenses]
 
 
-def get_licenses(session: Session) -> List[LicenseRow]:
-    db_licenses = session.execute(select(License)).scalars().all()
-    return [LicenseRow.from_orm(license) for license in db_licenses]
+async def remove_license(session: Session, license_name: str):
+    """
+    Remove the License from the database.
 
-
-def get_licenses_in_use(session: Session) -> List[LicenseInUseRow]:
-    db_licenses_in_use = session.execute(select(LicenseInUse)).scalars().all()
-    return [LicenseInUseRow.from_orm(license) for license in db_licenses_in_use]
-
-
-def get_licenses_in_use_from_name(session: Session, license_name: str) -> List[LicenseInUse]:
-    db_licenses_in_use = (
-        session.execute(select(LicenseInUse).where(LicenseInUse.license_name == license_name)).scalars().all()
+    Raises LicenseNotFound if the License does not exist.
+    """
+    query = await session.execute(select(License).where(License.name == license_name))
+    db_license = query.scalar_one_or_none()
+    enforce_defined(
+        db_license,
+        "License not found",
+        raise_exc_class=HTTPException,
+        exc_builder=lambda exc_class, msg: exc_class(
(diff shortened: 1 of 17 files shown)


lm-simulator/lm_simulator/crud.py
```diff
@@ -1,129 +1,155 @@
...
+    db_license = query.scalar_one_or_none()
+    enforce_defined(
+        db_license,
+        "License not found",
+        raise_exc_class=HTTPException,
+        exc_builder=lambda exc_class, msg: exc_class(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail="License not found",
+        ),
     )
```
dusktreader: You could combine these into a single expression since `enforce_defined()` returns the value it checks:
```suggestion
    db_license = enforce_defined(
        query.scalar_one_or_none()
        "License not found",
        raise_exc_class=HTTPException,
        exc_builder=lambda exc_class, msg: exc_class(
            status_code=status.HTTP_404_NOT_FOUND,
            detail="License not found",
        ),
    )
```

lm-simulator/lm_simulator/crud.py
```diff
@@ -1,129 +1,155 @@
...
-
-def _get_license_available(session: Session, license_name: str) -> LicenseRow:
-    license = session.execute(select(License).where(License.name == license_name)).scalars().first()
-    return license
 
+    with handle_errors(
+        "Can't remove License",
+        raise_exc_class=HTTPException,
+        exc_builder=lambda exc_class, msg: exc_class(
+            status_code=status.HTTP_400_BAD_REQUEST,
```
dusktreader: I feel like 400 is the wrong error code for this. Something went wrong in the database. That's a server error not a problem with the request. I would think this is a 500 or a 409.

lm-simulator/lm_simulator/crud.py
```diff
@@ -1,129 +1,155 @@
...
-    user_name: str,
-    quantity: int,
-    license_name: str,
-):
-    """Delete the license_in_use from the database.
+    with handle_errors(
+        "Can't create License In Use",
+        raise_exc_class=HTTPException,
+        exc_builder=lambda exc_class, msg: exc_class(
+            status_code=status.HTTP_400_BAD_REQUEST,
```
dusktreader: Same comment as above. I don't think 400 is the right error code for this.

lm-simulator/lm_simulator/crud.py
```diff
@@ -1,129 +1,155 @@
...
+        ),
+    )
 
-    session.execute(delete(License).where(License.name == license_name))
-    session.commit()
+    with handle_errors(
+        "Can't remove License In Use",
+        raise_exc_class=HTTPException,
+        exc_builder=lambda exc_class, msg: exc_class(
+            status_code=status.HTTP_400_BAD_REQUEST,
```
dusktreader: Maybe wrong status code? (see above)

lm-simulator/lm_simulator/schemas.py
```diff
@@ -1,4 +1,4 @@
-from typing import List, Optional
+from typing import List
```
dusktreader: Since we're on Python 3.12, we should use `list` in type hints instead of `List`.

lm-simulator/lm_simulator/schemas.py
```diff
@@ -22,19 +22,12 @@ class LicenseCreate(BaseModel):
...
+    model_config = ConfigDict(from_attributes=True)
 
     @model_validator(mode="after")
-    def in_use_validator(self) -> int:
-        """This validator is used to calculate the in_use value."""
-        self.in_use = 0
-        for license_in_use in self.licenses_in_use:
-            self.in_use += license_in_use.quantity
+    def calculate_in_use(self):
+        self.in_use = sum(license_in_use.quantity for license_in_use in self.licenses_in_use)
```
dusktreader: Won't this already be computed by the property `in_use` on the db model? Why re-implement here?

lm-simulator/tests/conftest.py
```diff
@@ -1,65 +1,153 @@
...
-@fixture(scope="session")
-def tables(engine):
-    Base.metadata.create_all(bind=engine)
-    yield
+@fixture
+async def backend_client(synth_session):
+    def override_get_session():
+        yield synth_session
+
+    subapp.dependency_overrides[get_session] = override_get_session
```
dusktreader, line 75: This is really cool! I didn't know about this

dusktreader (changes requested):
Nice work! Left some comments for you.

lm-simulator/lm_simulator/schemas.py
julianaklulo: That's a good point, I thought that the calculation had to happen in the schema as well.

lm-simulator/lm_simulator/crud.py
julianaklulo: Nice, I updated it!

lm-simulator/lm_simulator/crud.py
julianaklulo: Whe you can't create, it means that either the database operation went wrong or the license is already fully in use.
Is it possible to have a different treatment depending on the error returned by the database?

lm-simulator/lm_simulator/crud.py
julianaklulo: This one makes sense to be 500, I updated it.

lm-simulator/lm_simulator/crud.py
julianaklulo: I updated it to 500.

lm-simulator/lm_simulator/crud.py
julianaklulo: Since we're checking if there's enough licenses before inserting, that makes sense that an exception here means that something went wrong in the database. I updated it to 500.

dusktreader (approved):

PR #374, opened 2024-08-08
[ASP-4668] Add routes to Simulator API

julianaklulo:
#### What
Add a column for `License Server Type` in the License model.
Also add routes to filter by `License Server Type` and by license name.

#### Why
The scripts were filtering the licenses by name, expecting a license with a known name to distinguish the license server type. Now that the licenses in the database already have a type, it can be filtered in the API instead.

`Task`: https://jira.scania.com/browse/ASP-4668

---

#### Peer Review

### DIFF
--- a/lm-simulator/lm_simulator/crud.py
+++ b/lm-simulator/lm_simulator/crud.py
@@ -1,4 +1,4 @@
-from typing import List, Optional
+from typing import List
 
 from buzz import enforce_defined, handle_errors, require_condition
 from fastapi import HTTPException, status
@@ -60,7 +60,7 @@ async def list_licenses_by_server_type(session: Session, server_type: LicenseSer
     return [LicenseRow.model_validate(license) for license in db_licenses]
 
 
-async def read_license_by_name(session: Session, license_name: str) -> Optional[LicenseRow]:
+async def read_license_by_name(session: Session, license_name: str) -> LicenseRow:
     """
     Retrive the License in the database by name.
     """
--- a/lm-simulator/lm_simulator/crud.py
+++ b/lm-simulator/lm_simulator/crud.py
@@ -1,10 +1,11 @@
-from typing import List
+from typing import List, Optional
 
 from buzz import enforce_defined, handle_errors, require_condition
 from fastapi import HTTPException, status
 from sqlalchemy import select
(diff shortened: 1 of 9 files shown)


lm-simulator/lm_simulator/crud.py
```diff
@@ -50,6 +51,33 @@ async def list_licenses(session: Session) -> List[LicenseRow]:
...
+async def list_licenses_by_server_type(session: Session, server_type: LicenseServerType) -> List[LicenseRow]:
+    """
+    List all Licenses in the database by server type.
+    """
+    query = await session.execute(select(License).where(License.license_server_type == server_type))
+    db_licenses = query.scalars().all()
+    return [LicenseRow.model_validate(license) for license in db_licenses]
+
+
+async def read_license_by_name(session: Session, license_name: str) -> Optional[LicenseRow]:
```
fschuch: Is this `Optional` needed here?
This can be confusing for type checking, and I see the function uses `enforce_defined`.

lm-simulator/lm_simulator/models.py
```diff
@@ -17,6 +17,7 @@ class License(Base):
         lazy="selectin",
         cascade="all, delete-orphan",
     )
+    license_server_type: Mapped[str] = mapped_column(String, nullable=False)
```
fschuch, line 20: I see no migration file in this PR, is that ok?

fschuch (changes requested):
Good work. I left a couple of comments for you.

lm-simulator/lm_simulator/models.py
julianaklulo: The License Manager Simulator doesn't use alembic to manage migrations.
It was using a sqlite database until recently where I converted it to use PostgreSQL via docker-compose.
Since it's a project run only locally, I don't know if it's worth adding migrations for now.

lm-simulator/lm_simulator/crud.py
julianaklulo: That's a good point.
Before adding the validation, it was returning `None` if the license was not found.
But since it returns something or raises an exception  I'll remove the `Optional` then.
Thanks!

fschuch (approved):

PR #376, opened 2024-08-08
[ASP-4668] Split License Manager Simulator into two subprojects

julianaklulo:
#### What
Split the License Manager Simulator into two subprojects:
* `lm-simulator`: includes the fake binaries, templates and job application example
* `lm-simulator-api`: includes only the REST API used to populate the templates

#### Why
To best way to install the License Manager Simulator is by creating a Python package to distribute the scripts when installing the package. To do so we don't need the API to come bundled with the scripts, so splitting the project decouples both packages.

`Task`: https://jira.scania.com/browse/ASP-4668

---

#### Peer Review


.github/workflows/publish_on_tag.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
   build-publish:
     runs-on: ubuntu-latest
     strategy:
-      matrix: { sub_project: ['lm-agent', 'lm-api', 'lm-cli']}
+      matrix: { sub_project: ['lm-agent', 'lm-api', 'lm-cli', lm-simulator', 'lm-simulator-api'] }
```
dusktreader: Missing a quote here on `lm-simulator`

dusktreader (changes requested):

.github/workflows/publish_on_tag.yaml
julianaklulo: Nice catch, I fixed it!

dusktreader (approved):

PR #395, opened 2024-11-06
[ASP-5024]  Create lm-composed subproject

julianaklulo:
#### What
Create a `docker-compose` project to spin up an environment with all License Manager components working together.

#### Why
To facilitate integration testing.

`Task`: https://jira.scania.com/browse/ASP-5024

---

#### Peer Review


lm-api/Dockerfile
```diff
@@ -12,15 +12,17 @@ RUN poetry config virtualenvs.in-project true \
...
+
 COPY --from=builder /app /app
 
 ENV PATH="/app/.venv/bin:$PATH"
 
 COPY ./alembic /app/alembic
 COPY ./lm_api /app/lm_api
 
-CMD ["uvicorn", "lm_api.main:app", "--host", "0.0.0.0", "--port", "80", "--loop", "asyncio"]
+CMD ["uvicorn", "lm_api.main:app", "--host", "0.0.0.0", "--port", "7000", "--loop", "asyncio"]
```
dusktreader: Why did we need to change the port that's used inside the container?

lm-composed/etc/job_example.py
```diff
@@ -0,0 +1,83 @@
...
+    if headers is None:
+        headers = {"Content-Type": "application/json"}
+
+    request = urllib.request.Request(url=url, data=data, headers=headers, method=method)
+
+    try:
+        with urllib.request.urlopen(request) as response:
+            return response.getcode(), response.read().decode("utf-8")
+    except urllib.error.HTTPError as e:
+        print(f"HTTP Error: {e.code} - {e.reason}")
```
dusktreader: Just for correctness, these should print to `sys.stderr`.

lm-composed/etc/populate-lm-api.py
```diff
@@ -0,0 +1,78 @@
...
+        headers={"Authorization": f"Bearer {token}"}
+    ).json()
+
+    for row in rows:
+        _id = row["id"]
+        
+        httpx.delete(
+            f"{LM_AGENT_BACKEND_BASE_URL}/lm/{table_name}/{_id}",
+            headers={"Authorization": f"Bearer {token}"}
+        )
```
dusktreader: I think you should probably check the return code of these requests and throw an exception or exit with an error code if something goes wrong. Otherwise, it could be tricky to troubleshoot what's going wrong.

lm-composed/etc/seed-license-in-slurm.py
```diff
@@ -0,0 +1,43 @@
...
+import subprocess
+
+
+LICENSE_NAME = "test_product.test_feature"
+SERVER_TYPE = "flexlm"
+
+
+def main():
+    output = subprocess.run([
+        "scontrol",
```
dusktreader: It might be nice to shlex here to split up a single command string for readability.

lm-composed/etc/seed-license-in-slurm.py
```diff
@@ -0,0 +1,43 @@
...
+        f"{LICENSE_NAME}@{SERVER_TYPE}",
+    ], capture_output=True)
+
+    stdout = output.stdout.decode("utf-8")
+
+    if "test_product.test_feature@flexlm" in stdout:
+        return
+    else:
+        output = subprocess.run([
+            "sacctmgr",
```
dusktreader: See my shlex comment above. It really helps readability.

dusktreader (changes requested):
Awesome work! Left a few comments for you.

lm-api/Dockerfile
julianaklulo: I changed it during testing to avoid conflicting with another container but this change is not necessary so I reverted it.

lm-composed/etc/job_example.py
julianaklulo: Done!

lm-composed/etc/populate-lm-api.py
julianaklulo: Added error handling to this script, this way it will log and exit if the status code is not successful.

lm-composed/etc/seed-license-in-slurm.py
julianaklulo: Done!

lm-composed/etc/seed-license-in-slurm.py
julianaklulo: This one was a bit weird with `shlex`, but I was able to break it down to 2 lines.
What do you think?

dusktreader: It looks ok, but you will need to add an extra space at the beginning of the second line.

dusktreader (approved):

lm-composed/etc/seed-license-in-slurm.py
julianaklulo: Done!

PR #397, opened 2024-11-08
[ASP-5024] Remove DomainConfig from Armasec

julianaklulo:
#### What
Change the `Armasec` client to receive the configuration attributes directly instead of using `DomainConfig`.

#### Why
During the work on [ASP-5024] it was noted that `Armasec` was not logging even with `ARMASEC_DEBUG` set to `True`.
It was due to the way the client was being constructed.

`Task`: https://jira.scania.com/browse/ASP-5024

---

#### Peer Review

### DIFF
--- a/lm-api/lm_api/security.py
+++ b/lm-api/lm_api/security.py
@@ -8,7 +8,6 @@ import typing
 from typing_extensions import Self
 
 from armasec import Armasec, TokenPayload
-from armasec.schemas import DomainConfig
 from armasec.token_security import PermissionMode
 from fastapi import Depends
 from loguru import logger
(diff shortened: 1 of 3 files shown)


lm-api/lm_api/security.py
```diff
@@ -17,33 +16,11 @@
...
-                use_https=settings.ARMASEC_USE_HTTPS,
-            )
-        )
-    return domain_configs
-
-
-guard = Armasec(domain_configs=get_domain_configs())
+guard = Armasec(
+    domain=settings.ARMASEC_DOMAIN,
+    debug_logger=logger.debug if settings.ARMASEC_DEBUG else None,
```
matheushent, line 21: you could do something like [this](https://github.com/omnivector-solutions/vantage-api/blob/main/vantage-api/api/settings.py#L123) so the possibilites of logging are increased

julianaklulo: Sicne the API doesn't have a robust logging logic yet (it's in the backlog), this would be enough to get the debug logs from Armasec.
What would be the improvemenets from following this other approach?

matheushent: the ability to customize the logging level. This can be helpful when debugging a particular issue

dusktreader: I don't think you would ever want the armasec logs to be at any level other than "DEBUG". Myabe "TRACE". However, the way this is being set up right now mirrors how we are doing things in Jobbergate, and I think it's appropriate the way it is.

dusktreader (approved):

PR #426, opened 2025-04-03
chore(api): [ASP-6607] remove pytest-freezegun

julianaklulo:
The dependency `pytest-freezegun` was not being used by the `License Manager API` subproject, and since this package is not maintained anymore, it should be removed.


jamesbeedy (approved):
+1

PR #427, opened 2025-04-03
chore(lm-api,lm-simulator-api) [ASP-6607] remove version from docker-compose.yaml files

julianaklulo:
The `version` parameter on `docker-compose` was deprecated and it was raising warnings when running the projects.


jamesbeedy (approved):
+1

PR #471, opened 2025-10-16
fix: Update ruff configuration to properly format the code

julianaklulo:
This PR changes the `ruff` configuration to:
- Centralized `pyproject.toml` configuration file in repository's root
- Update each subproject Makefile to run:
    - Lint: `ruff check` and `ruff format --check`
    - Format: `ruff check --fix` and  `ruff format`
  
The current `ruff` configuration was not taking into consideration the formatting, just the linting.
Due to this, there were formatting errors in the projects that weren't captured before (fixed in this PR)


lm-agent/Makefile
```diff
@@ -19,18 +19,18 @@ mypy: install
 
 .PHONY: lint
 lint: install
-	poetry run ruff check ${PACKAGE_NAME}
-	poetry run ruff check tests
+	poetry run ruff check tests ${PACKAGE_NAME}
```
fschuch, line 22: It is a bit unclear to me that both `make lint` and `make format` run both format and lint verifications. I see arguments make one of them just verify, and the other try to fix, meybe we just need to clarify the names.

lm-agent/tests/parsing/test_flexlm.py
```diff
@@ -53,31 +53,31 @@ def test_parse_feature_line(line, result):
     "line,result",
     [
         (
-            "    user1 myserver.example.com /dev/tty (v62.2) (myserver.example.com/24200 12507), start Thu 10/29 8:09, 29 licenses",
+            "    user1 myserver.example.com /dev/tty (v62.2) (myserver.example.com/24200 12507), start Thu 10/29 8:09, 29 licenses",  # noqa: E501
```
fschuch: [optional] Configuring a per-file ignore can also work in this case.

lm-agent/tests/workload_managers/slurm/test_reservations.py
```diff
@@ -1,18 +1,20 @@
+# ruff: noqa: E501
```
fschuch: I think it works best setting a per-file ignore on the pyproject.toml level than on the file header.

pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[tool.ruff]
```
fschuch: Does it just out of the box? With no reference on the project's `pyproject.toml` file.

fschuch (changes requested):

lm-agent/tests/parsing/test_flexlm.py
julianaklulo: Good idea, the files that have more than one E501 are excluded now.

lm-agent/tests/workload_managers/slurm/test_reservations.py
julianaklulo: Done!

pyproject.toml
julianaklulo: According to the [docs](https://docs.astral.sh/ruff/configuration/#config-file-discovery):
`In locating the "closest" pyproject.toml file for a given path, Ruff ignores any pyproject.toml files that lack a [tool.ruff] section.`

lm-agent/Makefile
julianaklulo: That's a good point, the names are a bit confusing. To me `lint` includes all the checks for both styling and formatting, and `format` includes everything that changes the code.

Maybe having `make validate` and `make modify`?
Because `make check` can be confused with the `ruff` command `check`, and `make format` also runs into the same issue.

fschuch: I’m ok with validate and modify

julianaklulo: `make verify` might be clearer than `make validate` though

fschuch (approved):

PR #485, opened 2026-02-11
feat(lm-api): add metrics endpoint

julianaklulo:
Add a new `/lm/metrics` endpoint, using `Prometheus` client, to make license usage metrics available for `Prometheus` scrapers.

Note that the endpoint is protected by a new permission `METRICS_READ`, that needs to be set up in Keycloak for the user reading the data.


lm-api/lm_api/api/metrics/updater.py
```diff
@@ -0,0 +1,98 @@
...
+        task = self.task  # makes mypy happy
+        if task is None or task.done():
+            return
+
+        self._stop_event.set()
+        task.cancel()
+
+        try:
+            await task
+        except asyncio.CancelledError:
```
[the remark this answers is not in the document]
julianaklulo: Fixed

lm-api/lm_api/api/routes/metrics.py
fschuch, line 1: It looks good, I just miss some sort of mental model to illustrate what is going on, maybe on the module docstrinng, or router docstrings. I mean, looking at what the router file does and imports, it is not that evident it depends on a process started on main.py.

lm-api/lm_api/api/metrics/updater.py
```diff
@@ -0,0 +1,99 @@
...
+from lm_api.api.metrics.collector import collect_feature_metrics
+from lm_api.api.schemas.metrics import (
+    LICENSE_TOTAL,
+    LICENSE_USED,
+)
+from lm_api.config import settings
+from lm_api.database import engine_factory
+
+
+class MetricsManager:
```
fschuch: Please, add some reference for the protocol followed by this class, isn't there some base/abstract class from prometheus we can use here?

fschuch (changes requested):
Well done, I left some minor comments for you.

lm-api/lm_api/api/metrics/updater.py
julianaklulo: You're right, after a bit or research I found [Custom Collector](https://prometheus.github.io/client_python/collector/custom/) from `prometheus-client` that reduced a lot of the complexity that this updater was introducing to the API.
It only supports sync collection though, but considering that this is a read-only query, it shouldn't impact the performance.

lm-api/lm_api/api/routes/metrics.py
julianaklulo: I included more context in the docstrings.

fschuch (approved):

jamesbeedy (changes requested):
This looks great! Add a doc with an example or two of consuming the metrics endpoint and I'm +1.

jamesbeedy (approved):
+1

julianaklulo:
> This looks great! Add a doc with an example or two of consuming the metrics endpoint and I'm +1.

The docs are deserving a task of their own to get some care, I'll make sure to update it!

PR #564, opened 2026-07-16
feat(lm-agent): move agent off slurmctld

julianaklulo:
Rewrite the Prolog/Epilog scripts as lightweight `bash` scripts to remove the `lm-agent` dependency on the `slurmctld` node.


lm-prolog-epilog/acquire_oidc_token.sh
```diff
@@ -0,0 +1,116 @@
...
+    local token
+    token="$(json_value "$HTTP_BODY" "access_token")"
+    if [[ -z "$token" ]]; then
+        logger -t "lm-auth" "OIDC response did not contain an access_token"
+        return 1
+    fi
+
+    # Cache token with restrictive permissions
+    mkdir -p "$CACHE_DIR"
+    ( umask 077; echo -n "$token" > "$TOKEN_CACHE" )
```
fschuch: Just to tidy it, as suggested change, `> "$TOKEN_CACHE"` truncates the target first, so a partial write can corrupt the cache. Using a temp file plus atomic rename avoids exposing readers to a half-written token file. You can consider a atomic writing like:
```
tmp="$(mktemp "$CACHE_DIR/token.XXXXXX")" || exit 1
chmod 600 "$tmp" || exit 1
printf '%s' "$token" > "$tmp" || { rm -f "$tmp"; exit 1; }
mv -f "$tmp" "$TOKEN_CACHE"
```

fschuch (approved):
Nice work on this one.

lm-prolog-epilog/acquire_oidc_token.sh
julianaklulo: Nice catch, thanks!
I updated it to use the logic suggested.

